PR #1, opened 2022-01-27
First poc

benedeki:
First part of the code.
* Basic classes
* `NamingConventions` implementation - possible ways how to convert class name to database object name.

Please ignore the commented out code and `Enceladus` and `Examples` folders.


build.sbt
```diff
@@ -0,0 +1,45 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+ThisBuild / name         := "fa-db"
+ThisBuild / organization := "za.co.absa"
+
+lazy val scala211 = "2.11.12"
+lazy val scala212 = "2.12.12"
```
Zejnilovic, line 20: This library could in theory support 2.13 as well, right?

build.sbt
```diff
@@ -0,0 +1,45 @@
...
+
+ThisBuild / scalaVersion := scala211
+ThisBuild / crossScalaVersions := Seq(scala211, scala212)
+ThisBuild / publish := {}
```
AdrianOlosutean, line 24: I think this line should be removed

src/test/scala/za/co/absa/faDB/examples/ExampleSuite.scala
```diff
@@ -0,0 +1,28 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.faDB.examples
+
+import org.scalatest.funsuite.AnyFunSuite
+
+class ExampleSuite extends AnyFunSuite {
```
AdrianOlosutean, line 20: Why is this commented out?

build.sbt
benedeki: Depends on the lib(s) used to actually access the DB (Slick, Doobie,...) if they support it.
Otherwise I don't see reason why not. 😉

build.sbt
benedeki: I guess. I copped other projects mostly. 😄

src/test/scala/za/co/absa/faDB/examples/ExampleSuite.scala
benedeki: The whole examples folder will go It's just a remnant of the POC development process.
(I committed it with the first batch just not to loose the code before vacation. Will do a cleaning before release.)

build.sbt
Zejnilovic: IIRC this was made so it would not publish a root project. But as this project is just a root project without project definition, I agree with Adrian.

Then again cleaning of SBT can come later.

src/main/scala/za/co/absa/faDB/enceladus/DatasetSchema.scala
```diff
@@ -0,0 +1,124 @@
...
+  case class SchemaHead(schemaName: String, schemaLatestVersion: Int)
+
+  private implicit val SchemaHeadImplicit: GetResult[SchemaHead] = GetResult(r => {SchemaHead(r.<<, r.<<)})
+  private implicit val GetSchemaImplicit: GetResult[Schema] = GetResult(r => {
+    val status: Int = r.<<
+    val statusText: String = r.<<
+    if (status != 200) {
+      throw DBFailException(status, statusText)
+    }
+    Schema(r.<<, r.<<, r.<<, r.<<, r.<<, r.<<, r.<<, r.<<, r.<<, r.<<, r.<<, r.<<, r.<<)
```
Zejnilovic, line 75: This looks a bit funny. Are there other options?

src/main/scala/za/co/absa/faDB/enceladus/DatasetSchema.scala
```diff
@@ -0,0 +1,124 @@
...
+                    createdWhen: Timestamp,
+                    updatedBy: String,
+                    updatedWhen: Timestamp,
+                    lockedBy: Option[String],
+                    lockedWhen: Option[Timestamp],
+                    deletedBy: Option[String],
+                    deletedWhen: Option[Timestamp])
+  case class SchemaHead(schemaName: String, schemaLatestVersion: Int)
+
+  private implicit val SchemaHeadImplicit: GetResult[SchemaHead] = GetResult(r => {SchemaHead(r.<<, r.<<)})
```
Zejnilovic, line 68: This does not have a status and status text?

Zejnilovic (commented):
Posting to restart the PC

AdrianOlosutean:
Why is package naming camelCase? I would keep it acording to Scala standards

build.sbt
benedeki: SBT fixing do need to come. Planned as the third and last PR before first release.
Among other to make it a multi-package project - Core and Slick

src/main/scala/za/co/absa/faDB/enceladus/DatasetSchema.scala
benedeki: I hate it myself. But I haven't found an easy way to refer to the fields by name in Slick (remember it's ORM focused, the direct SQL is just a side-ability, this is what their docs suggest).
Although it should be possible, it seemed to me. I just didn't want to bother with its implementation for the moment. (A nice implicit enabling it, is another of the #plans 😄 )

src/main/resources/application.conf
```diff
@@ -0,0 +1,12 @@
...
+  dataSourceClass = "org.postgresql.ds.PGSimpleDataSource" //Simple datasource with no connection pooling. The connection pool has already been specified with HikariCP.
+  properties = {
+    serverName = "localhost"
+    portNumber = "5432"
+    databaseName = "menas_db"
+    user = "menas"
+    password = "menas"
+  }
+  numThreads = 10
+}
```
benedeki, line 12:
```suggestion
}

```

benedeki:
> Why is package naming camelCase? I would keep it acording to Scala standards

Me not paying enough attention when creating it. Thanks for noticing. Will fix in #3 to avoid too many conflicts (potentially)

AdrianOlosutean (approved):

PR #3, opened 2022-02-04
Separating db engine and core

benedeki:
* Separated the core and the Slick implementation
* Core is now independent and type-parametrized
* Enceladus moved to examples folder
* Deleted all kind unneeded code


benedeki:
This my be easier to review as the whole code rather than a PR with code compare. /methinks

dk1844 (approved):
LGTM (just read the code). To the degree that I have understood it, I see nothing wrong.

src/main/scala/za/co/absa/fadb/naming_conventions/lettersCase/LettersCase.scala
```diff
@@ -13,7 +13,7 @@
  * limitations under the License.
  */
 
-package za.co.absa.faDB.namingConventions.lettersCase
+package za.co.absa.fadb.naming_conventions.lettersCase
```
AdrianOlosutean, line 16: Probably `lettersCase` package should also be renamed

dk1844 (approved):
LGTM (just read the code)

AdrianOlosutean (approved):
Looks good

PR #52, opened 2023-07-06
#51: Simplify usage by offering combined classes

### LINKED ISSUE #51: Simplify usage by offering combined classes
## Background
When using Fa-Db a `DbFunction` descendant class and mix-in trait(s) have to be declared for each function. That's unnecessarily tedious.

## Feature
Simplify usage by offering classes that already combine `DBFunction` descendants combined with mix-in traits.

benedeki:
* `DBMultipleResultFunction` added
* `DBMultipleResultFunctionWithStatusSupport` added
* `DBOptionalResultFunction` added
* `DBOptionalResultFunctionWithStatusSupport` added
* `DBSingleResultFunction` added
* `DBSingleResultFunctionWithStatusSupport`

Closes #51

### DIFF
--- a/slick/src/main/scala/za/co/absa/fadb/slick/pg/DBFunction.scala
+++ b/slick/src/main/scala/za/co/absa/fadb/slick/pg/DBFunction.scala
@@ -1,0 +1,278 @@
+/*
+ * Copyright 2023 ABSA Group Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.fadb.slick.pg
+
+import za.co.absa.fadb.DBSchema
+import za.co.absa.fadb.slick.{SlickFunction, SlickFunctionWithStatusSupport, SlickPgEngine}
+import za.co.absa.fadb.DBFunction.{DBMultipleResultFunction => CoreDBMultipleResultFunction, DBOptionalResultFunction => CoreDBOptionalResultFunction, DBSingleResultFunction => CoreDBSingleResultFunction}
+
+object DBFunction {
+  /**
+    * Represents a function returning a set (in DB sense) of rows
+    *
+    * @param functionNameOverride - in case the class name would not match the database function name, this gives the
+    *                             possibility of override
+    * @param schema               - the schema the function belongs into
+    * @param dBEngine             - the database engine that is supposed to execute the function (presumably contains
+    *                             connection to the database
+    * @tparam I - the type covering the input fields of the database function
+    * @tparam R - the type covering the returned fields from the database function
+    */
+  abstract class DBMultipleResultFunction[I, R](functionNameOverride: Option[String] = None)
+                                               (implicit schema: DBSchema, dBEngine: SlickPgEngine)
+    extends CoreDBMultipleResultFunction[I, R, SlickPgEngine](functionNameOverride)
+      with SlickFunction[I, R] {
+
+    def this(functionNameOverride: String)
+            (implicit schema: DBSchema, dBEngine: SlickPgEngine) = {
+      this(Option(functionNameOverride))(schema, dBEngine)
+    }
+
+    def this(schema: DBSchema, functionNameOverride: String)
+            (implicit dBEngine: SlickPgEngine) = {
+      this(Option(functionNameOverride))(schema, dBEngine)
+    }
+
+    def this(schema: DBSchema)
+            (implicit dBEngine: SlickPgEngine) = {
+      this(None)(schema, dBEngine)
+    }
+
+    def this(dBEngine: SlickPgEngine, functionNameOverride: String)
+            (implicit schema: DBSchema) = {
+      this(Option(functionNameOverride))(schema, dBEngine)
+    }
+
+    def this(dBEngine: SlickPgEngine)
+            (implicit schema: DBSchema) = {
+      this(None)(schema, dBEngine)
+    }
+  }
+
+  /**
+    * Represents a function returning exactly one record
+    *
+    * @param functionNameOverride - in case the class name would not match the database function name, this gives the
+    *                             possibility of override
+    * @param schema               - the schema the function belongs into
+    * @param dBEngine             - the database engine that is supposed to execute the function (presumably contains
+    *                             connection to the database
+    * @tparam I - the type covering the input fields of the database function
+    * @tparam R - the type covering the returned fields from the database function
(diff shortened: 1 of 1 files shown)


slick/src/main/scala/za/co/absa/fadb/slick/pg/DBFunction.scala
```diff
@@ -0,0 +1,278 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.fadb.slick.pg
+
+import za.co.absa.fadb.DBSchema
+import za.co.absa.fadb.slick.{SlickFunction, SlickFunctionWithStatusSupport, SlickPgEngine}
+import za.co.absa.fadb.DBFunction.{DBMultipleResultFunction => CoreDBMultipleResultFunction, DBOptionalResultFunction => CoreDBOptionalResultFunction, DBSingleResultFunction => CoreDBSingleResultFunction}
```
lsulak, line 21:
```suggestion
import za.co.absa.fadb.DBFunction.{
  DBMultipleResultFunction => CoreDBMultipleResultFunction, 
  DBOptionalResultFunction => CoreDBOptionalResultFunction, 
  DBSingleResultFunction => CoreDBSingleResultFunction
}
```

very long line, this formatting looks better

slick/src/main/scala/za/co/absa/fadb/slick/pg/DBFunction.scala
```diff
@@ -0,0 +1,278 @@
...
+    *                             connection to the database
+    * @tparam I - the type covering the input fields of the database function
+    * @tparam R - the type covering the returned fields from the database function
+    */
+  abstract class DBMultipleResultFunction[I, R](functionNameOverride: Option[String] = None)
+                                               (implicit schema: DBSchema, dBEngine: SlickPgEngine)
+    extends CoreDBMultipleResultFunction[I, R, SlickPgEngine](functionNameOverride)
+      with SlickFunction[I, R] {
+
+    def this(functionNameOverride: String)
```
lsulak, line 40: Are all these constructors really needed? 

The only difference between `DBMultipleResultFunction` and `CoreDBMultipleResultFunction` (I know that's just an alias but let's use that for the sake of being able to make a distinction) is the engine, and you are passing the `SlickPgEngine` into appropriate place when you are extending it, so it should be informed about the concrete type; thus you shouldn't need to reimplement the constructors...? Am I missing something?

slick/src/main/scala/za/co/absa/fadb/slick/pg/DBFunction.scala
```diff
@@ -0,0 +1,278 @@
...
+    }
+
+    def this(dBEngine: SlickPgEngine)
+            (implicit schema: DBSchema) = {
+      this(None)(schema, dBEngine)
+    }
+  }
+
+  /**
+    * Represents a function returning exactly one record
```
lsulak, line 67: All this class documentation(s) are duplicated - the base classes contain them, and it's almost or nearly the same text. `@see` should work, please try it: https://stackoverflow.com/questions/53850885/how-to-use-see-scaladoc

```suggestion
    @see [[za.co.absa.fadb.DBFunction.DBSingleResultFunction]]
```

lsulak (commented):
LGTM although I think that there is a big potential for a lot of code deletion. Please see my comments

slick/src/main/scala/za/co/absa/fadb/slick/pg/DBFunction.scala
benedeki: Interesting, I thought I did it the suggested way...

slick/src/main/scala/za/co/absa/fadb/slick/pg/DBFunction.scala
benedeki: Good spot!

slick/src/main/scala/za/co/absa/fadb/slick/pg/DBFunction.scala
```diff
@@ -0,0 +1,282 @@
...
+    }
+
+    def this(dBEngine: SlickPgEngine)
+            (implicit schema: DBSchema) = {
+      this(None)(schema, dBEngine)
+    }
+  }
+
+  /**
+    @see [[za.co.absa.fadb.DBFunction.DBSingleResultFunction]]
```
lsulak, line 71: btw I would be as harsh as this - let's remove all the code doc (these params) and have a "one liner" containing just this @see for the whole class

benedeki:
Doc is suddenly failing - on a class that has not been changed. But found other (naming) issues, that I want to address.

slick/src/main/scala/za/co/absa/fadb/slick/pg/DBFunction.scala
lsulak: Okay as we talked about elsewhere, auxiliary constructors are not inherited in Scala, so my idea cannot be applied here. Sorry.

salamonpavel:
The topic of this PR has been addressed in PR #106. I believe we can close this one.

slick/src/main/scala/za/co/absa/fadb/slick/pg/DBFunction.scala
```diff
@@ -0,0 +1,243 @@
...
+import za.co.absa.fadb.slick.{SlickFunction, SlickFunctionWithStatusSupport, SlickPgEngine}
+import za.co.absa.fadb.DBFunction.{
+  DBMultipleResultFunction => CoreDBMultipleResultFunction,
+  DBOptionalResultFunction => CoreDBOptionalResultFunction,
+  DBSingleResultFunction => CoreDBSingleResultFunction
+}
+
+object DBFunction {
+  /**
+    * Combines [[za.co.absa.fadb.DBFunction.DBMultipleResultFunction DBMultipleResultFunction]] with Slick and Postgres support
```
lsulak, line 29:
```suggestion
    * Combines [[za.co.absa.fadb.DBFunction.DBMultipleResultFunction]] with Slick and Postgres support
```

❓

slick/src/main/scala/za/co/absa/fadb/slick/pg/DBFunction.scala
```diff
@@ -0,0 +1,243 @@
...
+    }
+
+    def this(dBEngine: SlickPgEngine)
+            (implicit schema: DBSchema) = {
+      this(None)(schema, dBEngine)
+    }
+  }
+
+  /**
+    * Combines [[za.co.absa.fadb.DBFunction.DBSingleResultFunction DBSingleResultFunction]] with Slick and Postgres support
```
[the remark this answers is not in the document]
lsulak: and so on for the rest of them below

lsulak (commented):
I don't see much value in this PR to be fair :( I think that users should combine whatever they want and that this code brings only very little convenience, but the drawback is that it's a code and each code must be maintained, tested, understood, and documented.

Besides, I think that something like this was already implemented recently and thus I propose to close this PR. For Slick, see this: https://github.com/AbsaOSS/fa-db/pull/106/files#diff-b585ca0e03975d3e0dbdbedca5e6a38f067b043218992ebf9e20dc03b98af2c4 and Doobie has its equivalent in relevant files

Issue #53, opened 2023-09-18
Create a deployment script


benedeki:
## Background
Deploying SQL files is tedious and some can be easily skipped.

## Feature
Create a deployment script that will deploy database object based on provided SQL files

PR #54, opened 2023-09-18
#53: Create a deployment script

benedeki:
* deployment script

### DIFF
--- a/scripts/deployment/deploy.py
+++ b/scripts/deployment/deploy.py
@@ -1,0 +1,182 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+import dataclasses
+import argparse
+import logging
+import os
+import psycopg2
+import copy
+
+
+@dataclasses.dataclass
+class PostgresDBConn:
+    """This dataclass contains all information related to making a connection to Postgres DB."""
+    username: str
+    password: str
+    host: str
+    database: str
+    port: int
+
+
+def parse_args() -> argparse.Namespace:
+    """CLI args parsing function."""
+    parser = argparse.ArgumentParser(
+        description="Deploys structures to DB. (script version: 1.0)",
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter,
+    )
+    parser.add_argument(
+        "-ph", "--host",
+        help="database server host (default: \"localhost\")",
+        default="localhost",
+    )
+    parser.add_argument(
+        "-p", "--port",
+        help="database server port (default: \"5432\")",
+        default="5432",
+    )
+    parser.add_argument(
+        "-d", "--dbname",
+        help="database name to connect to (default: \"ursa_unify_db\")",
+        required=True,
+    )
+    parser.add_argument(
+        "-U", "--username",
+        help="database user name, should be a high privileged account (default: \"postgres\")",
+        default="postgres",
+    )
+    parser.add_argument(
+        "-W", "--password",
+        help="database user password",
+        required=True,
+    )
+    parser.add_argument(
+        "-dir", "--dir",
+        help="the directory of database source files (default: current directory)",
+        default=os.getcwd()
+    )
+    parser.add_argument(
+        "--create-db",
+        action="store_true",
+        help="creates the target database (runs the scripts in the source root starting with '00_', which is/are expected creating the db",
+    )
+
+    return parser.parse_args()
+
+
+def execute_sql(conn_config: PostgresDBConn, sql: str):
+    conn = psycopg2.connect(
+        database=conn_config.database,
+        user=conn_config.username,
+        password=conn_config.password,
+        host=conn_config.host,
+        port=conn_config.port
+    )
+
+    conn.autocommit = True
+    cursor = conn.cursor()
+
+    cursor.execute(sql)
+
+    conn.commit()
+    conn.close()
+
+
+def read_file(file_name: str) -> str:
+    logging.debug(f"    - reading file `{file_name}`")
+    file = open(file_name, "r")
+    result = file.read()
+    file.close()
+    return result
+
+
+def ensure_trailing_slash(path: str) -> str:
+    if path.endswith('/'):
+        return path
+    else:
+        return path + '/'
+
+
+def process_dir(directory: str, conn_config: PostgresDBConn, create_db: bool):
+    logging.info(f"Picking up source files from directory `{directory}`")
+    public_schema = "public"
+    root = next(os.walk(directory), (None, [], []))
+    schemas = list(root[1])
+    files = list(filter(lambda fn: fn.endswith(".ddl"), root[2]))
+
+    # process root files
+    database_creation_sqls = []
+    init_sqls = []
+    for filename in files:
+        if filename.startswith("00_"):
+            database_creation_sqls.append(filename)
+        else:
+            init_sqls.append(filename)
+    # process schemas
+    schemas_sqls = []
+    if public_schema in schemas:
+        # public folder has to go first
+        schemas.remove(public_schema)
+        schemas_sqls += process_schema(directory, public_schema, False)
+
+    for schema in schemas:
+        schemas_sqls += process_schema(directory, schema, True)
+
+    # execute the collected Sqls
+    if (len(database_creation_sqls) > 0) and create_db:
+        logging.info("Creating database")
+        db_conn_config = copy.copy(conn_config)
+        db_conn_config.database = "postgres"
+        sql = "\n".join(map(read_file, database_creation_sqls))
+        execute_sql(db_conn_config, sql)
+
+    if len(init_sqls) > 0:
+        logging.info("Initializing the database")
+        sql = "\n".join(map(read_file, init_sqls))
+        execute_sql(conn_config, sql)
+    if len(schemas_sqls) > 0:
+        logging.info("Populating the schemas")
+        sql = "\n".join(schemas_sqls)
+        execute_sql(conn_config, sql)
+
+
+def process_schema(base_dir: str,  schema_name: str, expect_schema_creation: bool) -> list[str]:
+    logging.info(f"  - schema '{schema_name}'")
+    schema_dir = ensure_trailing_slash(base_dir + schema_name)
+    schema_creation = []
+    functions = []
+    tables = []
+    has_schema_creation = False
+    files = os.listdir(schema_dir)
+    files.sort()
+    for input_file in files:
+        if input_file == "_.ddl":
+            schema_creation = [read_file(schema_dir + input_file)]
+            has_schema_creation = True
+        elif input_file.endswith(".sql"):
+            functions.append(read_file(schema_dir + input_file))
+        elif input_file.endswith(".ddl"):
+            tables.append(read_file(schema_dir + input_file))
+
+    if expect_schema_creation and (not has_schema_creation):
+        logging.warning(f"No schema creation found on path `{schema_dir}`")
+
+    return schema_creation + functions + tables
+
+
+if __name__ == '__main__':
+    logging.basicConfig(
+        level=logging.DEBUG, format="%(asctime)s,%(msecs)03d %(levelname)-8s [%(filename)s:%(lineno)s] %(message)s",
+    )
+
+    parsed_args = parse_args()
+
+    db_conn_details = PostgresDBConn(
+        username=parsed_args.username,
+        password=parsed_args.password,
+        host=parsed_args.host,
+        database=parsed_args.dbname,
+        port=parsed_args.port
+    )
+
+    process_dir(ensure_trailing_slash(parsed_args.dir), db_conn_details, parsed_args.create_db)


scripts/deployment/deploy.py
```diff
@@ -0,0 +1,182 @@
...
+
+
+def parse_args() -> argparse.Namespace:
+    """CLI args parsing function."""
+    parser = argparse.ArgumentParser(
+        description="Deploys structures to DB. (script version: 1.0)",
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter,
+    )
+    parser.add_argument(
+        "-ph", "--host",
```
lsulak, line 29: why not just `-h` ? I think that's more standard across the generally available tools

scripts/deployment/deploy.py
```diff
@@ -0,0 +1,182 @@
...
+        help="database server host (default: \"localhost\")",
+        default="localhost",
+    )
+    parser.add_argument(
+        "-p", "--port",
+        help="database server port (default: \"5432\")",
+        default="5432",
+    )
```
lsulak, line 37: I would expect something like this instead:

```
    parser.add_argument(
        "-p", "--port",
        help="database server port (default: 5432)",
        action="store",
        default=5432,
        type=int,
    )
```

so that it's int automatically

scripts/deployment/deploy.py
```diff
@@ -0,0 +1,182 @@
...
+    parser.add_argument(
+        "-dir", "--dir",
+        help="the directory of database source files (default: current directory)",
+        default=os.getcwd()
+    )
+    parser.add_argument(
+        "--create-db",
+        action="store_true",
+        help="creates the target database (runs the scripts in the source root starting with '00_', which is/are expected creating the db",
```
lsulak, line 61:
```suggestion
        help="creates the target database (runs the scripts in the source root starting with '00_', which is/are "
             "expected to create the db",
```

scripts/deployment/deploy.py
```diff
@@ -0,0 +1,182 @@
...
+    password: str
+    host: str
+    database: str
+    port: int
+
+
+def parse_args() -> argparse.Namespace:
+    """CLI args parsing function."""
+    parser = argparse.ArgumentParser(
+        description="Deploys structures to DB. (script version: 1.0)",
```
lsulak, line 25: not sure why information about the script version is needed but I would just not mention it

scripts/deployment/deploy.py
```diff
@@ -0,0 +1,182 @@
...
+    result = file.read()
+    file.close()
+    return result
+
+
+def ensure_trailing_slash(path: str) -> str:
+    if path.endswith('/'):
+        return path
+    else:
+        return path + '/'
```
lsulak, line 97: Can be a one liner:

`return path if path.endswith('/') else path + '/'`

scripts/deployment/README.md
```diff
@@ -0,0 +1,13 @@
+# Deployment script
+
+This _Python_ script is there to help deploy DB objects at a clear Postgres database or database server.
+
+Expectations are that thes ource files are placed in a directory structure of following properties
```
lsulak, line 5:
```suggestion
Expectations are that the source files are placed in a directory structure of following properties
```

scripts/deployment/README.md
```diff
@@ -0,0 +1,13 @@
...
+* database creation script is stored in source root, starts with `00_` and has the extension of `.ddl` (those files are skipped unless a database creation switch is provided to the deployment script)
+* other required systemic changes (users, extensions) are stored in files in root, their extension is `.ddl` and are processed alphabetically
+* it's recommended these to be written in re-exutable way, of they affet the whole server not just the database
+* objects of schemas are in folders (convention: folder name equals schema name)
+* schema creation SQL is stored in file `_.ddl`
+* * DB functions are stored in files with `.sql` extension (conventions: file name equals function name)
```
lsulak, line 11:
```suggestion
* DB functions are stored in files with `.sql` extension (conventions: file name equals function name)
```

❓

scripts/deployment/deploy.py
```diff
@@ -0,0 +1,182 @@
...
+        port=conn_config.port
+    )
+
+    conn.autocommit = True
+    cursor = conn.cursor()
+
+    cursor.execute(sql)
+
+    conn.commit()
+    conn.close()
```
lsulak, line 82: I don't know whether the connection will be properly closed on an error in the query or unexpected interruption...like someone stops the script in the middle.

The way I overcame this problem in the migration script DynamoDb -> Postgres was to implement this functionality as a context manager

scripts/deployment/deploy.py
```diff
@@ -0,0 +1,182 @@
...
+
+
+def read_file(file_name: str) -> str:
+    logging.debug(f"    - reading file `{file_name}`")
+    file = open(file_name, "r")
+    result = file.read()
+    file.close()
```
lsulak, line 89: Let's use context manager for opening and closing files - it will be handled properly even in case of a failure.

Somethig like 

```
with open(file_name, "r") as file:
    return file.read()
```

scripts/deployment/README.md
```diff
@@ -0,0 +1,13 @@
+# Deployment script
```
lsulak, line 1: Consider adding a brief section about installation / dependencies / run, and adding a text file with dependencies, usually called `requirements.txt` with content like this:

```
psycopg2==2.9.6
```

scripts/deployment/deploy.py
```diff
@@ -0,0 +1,182 @@
...
+    else:
+        return path + '/'
+
+
+def process_dir(directory: str, conn_config: PostgresDBConn, create_db: bool):
+    logging.info(f"Picking up source files from directory `{directory}`")
+    public_schema = "public"
+    root = next(os.walk(directory), (None, [], []))
+    schemas = list(root[1])
+    files = list(filter(lambda fn: fn.endswith(".ddl"), root[2]))
```
lsulak, line 105: I would even extract these ".ddl" strings and "00_" strings into a global constant, and naming it something like 

`DDL_FILES = ".ddl"`
`DB_CREATION_FILES = "00_"`

scripts/deployment/deploy.py
```diff
@@ -0,0 +1,182 @@
...
+
+    for schema in schemas:
+        schemas_sqls += process_schema(directory, schema, True)
+
+    # execute the collected Sqls
+    if (len(database_creation_sqls) > 0) and create_db:
+        logging.info("Creating database")
+        db_conn_config = copy.copy(conn_config)
+        db_conn_config.database = "postgres"
+        sql = "\n".join(map(read_file, database_creation_sqls))
```
lsulak, line 130: the intent is quite understandable by reading this line although the var name `sql` feels a bit unclear

scripts/deployment/deploy.py
```diff
@@ -0,0 +1,182 @@
...
+        logging.info("Initializing the database")
+        sql = "\n".join(map(read_file, init_sqls))
+        execute_sql(conn_config, sql)
+    if len(schemas_sqls) > 0:
+        logging.info("Populating the schemas")
+        sql = "\n".join(schemas_sqls)
+        execute_sql(conn_config, sql)
+
+
+def process_schema(base_dir: str,  schema_name: str, expect_schema_creation: bool) -> list[str]:
```
lsulak, line 143:
```suggestion
def process_schema(base_dir: str, schema_name: str, expect_schema_creation: bool) -> list[str]:
```

scripts/deployment/deploy.py
```diff
@@ -0,0 +1,182 @@
...
+    for input_file in files:
+        if input_file == "_.ddl":
+            schema_creation = [read_file(schema_dir + input_file)]
+            has_schema_creation = True
+        elif input_file.endswith(".sql"):
+            functions.append(read_file(schema_dir + input_file))
+        elif input_file.endswith(".ddl"):
+            tables.append(read_file(schema_dir + input_file))
+
+    if expect_schema_creation and (not has_schema_creation):
```
lsulak, line 161:
```suggestion
    if expect_schema_creation and not has_schema_creation:
```

scripts/deployment/deploy.py
benedeki: It's a stupid script. I would like to have some version somewhere. And if I am having it, why not print it with help?

scripts/deployment/deploy.py
```diff
@@ -0,0 +1,182 @@
...
+    root = next(os.walk(directory), (None, [], []))
+    schemas = list(root[1])
+    files = list(filter(lambda fn: fn.endswith(".ddl"), root[2]))
+
+    # process root files
+    database_creation_sqls = []
+    init_sqls = []
+    for filename in files:
+        if filename.startswith("00_"):
+            database_creation_sqls.append(filename)
```
miroslavpojer, line 112:
```suggestion
            database_creation_sqls.append(os.path.join(directory, filename))
```

scripts/deployment/deploy.py
```diff
@@ -0,0 +1,182 @@
...
+        else:
+            init_sqls.append(filename)
```
miroslavpojer, line 114:
```suggestion
            init_sqls.append(os.path.join(directory, filename))
```

scripts/deployment/deploy.py
lsulak: But I would rather remove this function completely and on places where you need to use it - and the only real & mandatory presence of '/' is when you are concatenating paths, I would use something like this: `os.path.join(x, y)` - this is much better and multi-platform

scripts/deployment/deploy.py
```diff
@@ -0,0 +1,182 @@
...
+        db_conn_config.database = "postgres"
+        sql = "\n".join(map(read_file, database_creation_sqls))
+        execute_sql(db_conn_config, sql)
+
+    if len(init_sqls) > 0:
+        logging.info("Initializing the database")
+        sql = "\n".join(map(read_file, init_sqls))
+        execute_sql(conn_config, sql)
+    if len(schemas_sqls) > 0:
+        logging.info("Populating the schemas")
```
miroslavpojer, line 138: This is the last log line in console output. After verbose output from another logical section it looks like the something happen and no more output.
Add some confirmation about start - end or list of executed sql file to keep same format of output.

miroslavpojer (commented):
- [ ] pull
- [ ] code review
- [ ] run

scripts/deployment/deploy.py
```diff
@@ -0,0 +1,182 @@
...
+    schema_dir = ensure_trailing_slash(base_dir + schema_name)
+    schema_creation = []
+    functions = []
+    tables = []
+    has_schema_creation = False
+    files = os.listdir(schema_dir)
+    files.sort()
+    for input_file in files:
+        if input_file == "_.ddl":
+            schema_creation = [read_file(schema_dir + input_file)]
```
lsulak, line 154:
```suggestion
            schema_creation = [read_file(os.path.join(schema_dir, input_file))]
```

also on other places in this func

scripts/deployment/deploy.py
```diff
@@ -0,0 +1,182 @@
...
+        logging.info("Populating the schemas")
+        sql = "\n".join(schemas_sqls)
+        execute_sql(conn_config, sql)
+
+
+def process_schema(base_dir: str,  schema_name: str, expect_schema_creation: bool) -> list[str]:
+    logging.info(f"  - schema '{schema_name}'")
+    schema_dir = ensure_trailing_slash(base_dir + schema_name)
+    schema_creation = []
+    functions = []
```
lsulak, line 147: A reader of this code can't know whether this is a list of function names, SQL code representing function implementations, or list of function files, or something else. Much more descriptive name of this (and `tables` ) var would be good to have.

Like, `functions_as_sql` or something like that

scripts/deployment/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+This _Python_ script is there to help deploy DB objects at a clear Postgres database or database server.
```
lsulak, line 3: A very simple example of a structure that this script is requiring is absolutely needed I would say

scripts/deployment/deploy.py
lsulak: `action="store"` just stores the argument’s value, and it's the default value for `add_argument` function, so you can omit it

scripts/deployment/deploy.py
lsulak: except for rare circumstances, splitting anything so that line is not too long is a good thing in my view :) 

and in terms of finding the text - I understand this argument, but I never had problems with it even in codebase with dozens of lines of code. If I cannot find some message then it means that it was split or constructed dynamically, so I'm just trying to find some substring

scripts/deployment/deploy.py
```diff
@@ -0,0 +1,182 @@
...
+    files = os.listdir(schema_dir)
+    files.sort()
+    for input_file in files:
+        if input_file == "_.ddl":
+            schema_creation = [read_file(schema_dir + input_file)]
+            has_schema_creation = True
+        elif input_file.endswith(".sql"):
+            functions.append(read_file(schema_dir + input_file))
+        elif input_file.endswith(".ddl"):
+            tables.append(read_file(schema_dir + input_file))
```
lsulak, line 159: potential optimization could be to only get the file names in this function, and then the loading of their content could be done in the same way as on line `"\n".join(map(read_file, init_sqls))`, so it would be consistent and a bit more memory efficient

scripts/deployment/deploy.py
```diff
@@ -0,0 +1,182 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
```
lsulak, line 2: consider renaming this file from `deploy.py` to `deploy_pg;py`

scripts/deployment/deploy.py
lsulak: I'll leave it up to you, but versioning individual scripts (for which purpose?) seems odd, especially since we are versioning the whole repo, i.e. we are not versioning individual files or modules, just the whole codebase 🤷‍♂️

scripts/deployment/deploy.py
benedeki: Because `-h` is already taken by _argparse_ (for help display). I tried, have it that way originally.

scripts/deployment/deploy.py
benedeki: Accepted :-)

scripts/deployment/deploy.py
```diff
@@ -0,0 +1,182 @@
...
+
+def process_dir(directory: str, conn_config: PostgresDBConn, create_db: bool):
+    logging.info(f"Picking up source files from directory `{directory}`")
+    public_schema = "public"
+    root = next(os.walk(directory), (None, [], []))
+    schemas = list(root[1])
+    files = list(filter(lambda fn: fn.endswith(".ddl"), root[2]))
+
+    # process root files
+    database_creation_sqls = []
```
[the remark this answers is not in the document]
benedeki: Renamed for better, I hope

scripts/deployment/deploy.py
```diff
@@ -0,0 +1,182 @@
...
+    init_sqls = []
```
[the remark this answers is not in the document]
benedeki: Also renamed

scripts/deployment/deploy.py
benedeki: Renamed too.

scripts/deployment/deploy.py
benedeki: More consitent maybe.
More memory efficient why?

lsulak: > More memory efficient why?

because now it has to be hold in memory from this point until it get's garbage collected which is probably after some point of running the sql statements.

If read only in that join function, then the memory requirements are the same but it will be held in memory probably for shorter time. Not a big deal in this scenario I know

scripts/deployment/deploy_pg.py
```diff
@@ -0,0 +1,184 @@
...
+    )
+    parser.add_argument(
+        "-ph", "--host",
+        help="database server host (default: \"localhost\")",
+        default="localhost",
+    )
+    parser.add_argument(
+        "-p", "--port",
+        help="database server port (default: \"5432\")",
+        default="5432",
```
lsulak, line 43:
```suggestion
        default=5432,
```

perhaps it would be converted automatically anyway though. But Python is a magical place where anything is allowed and that makes software very vulnerable to mistakes

lsulak (commented):
I've tried this script on AUL DB files, and it worked when. However, repetitive runs don't work - generally speaking there is `CREATE TABLE X` and not `CREATE TABLE IF NOT EXISTS X` - detecting changes on tables would be complicated for this script and its intended (temporary) use = so we won't be able to replace / update tables with this script but that's okay - besides, functions will be replaced, that's good. 

But I feel that this script shouldn't fail and we should be able to replace the implementation of DB functions with it - but now the script fails saying that a table already exists - we could capture this exception `psycopg2.errors.DuplicateTable` maybe? Or even better, perhaps we could ignore `Initializing the database...` step and it could be driven by some CLI param.?

So I propose to have this:
* CREATE DB - no by default, but can be enabled
* INIT DB - no by default, but can be enabled - this would need one more CLI parameter 
* Populate schemas - yes by default, cannot be disabled

```
    parser.add_argument(
        "--init-db",
        action="store_true",
        help="initializes the target database (runs the scripts ending with '.ddl' that don't start with '00_'"
    )
```

scripts/deployment/deploy_pg.py
```diff
@@ -0,0 +1,184 @@
...
+        conn.close()
+
+
+def read_file(file_name: str) -> str:
+    logging.debug(f"    - reading file `{file_name}`")
+    with open(file_name, "r") as file:
+        return file.read()
+
+
+def process_dir(directory: str, conn_config: PostgresDBConn, create_db: bool) -> None:
```
lsulak, line 101: i still think that this function is a bit too long / doing too much, I would definitely split the processing of those 3 file categories into 3 separate functions - but I don't insist since this is a temporary script, still very useful.

An obvious improvement would be also to check whether the CLI DB name and DB name from the SQL files are matching, or to have the SQL files as Jinja2 templates and insert all necessary values dynamically. But that would be probably too much for this script & now

lsulak (commented):
- [x]  code reviewed
- [x]  pulled
- [x]  built
- [x]  ran on my machine against my local PG db

found a few issues and I would love one thing to implement (repetitive run support), otherwise I'm happy to approve

scripts/deployment/deploy_pg.py
benedeki: You are exactly right in all what you write.
And also have good reasoning why not to do it. Hopefully we will have good deployment tool, after...

miroslavpojer (commented):
- [x] tested

lsulak:
> I've tried this script on AUL DB files, and it worked when. However, repetitive runs don't work - generally speaking there is `CREATE TABLE X` and not `CREATE TABLE IF NOT EXISTS X` - detecting changes on tables would be complicated for this script and its intended (temporary) use = so we won't be able to replace / update tables with this script but that's okay - besides, functions will be replaced, that's good.
> 
> But I feel that this script shouldn't fail and we should be able to replace the implementation of DB functions with it - but now the script fails saying that a table already exists - we could capture this exception `psycopg2.errors.DuplicateTable` maybe? Or even better, perhaps we could ignore `Initializing the database...` step and it could be driven by some CLI param.?
> 
> So I propose to have this:
> 
> * CREATE DB - no by default, but can be enabled
> * INIT DB - no by default, but can be enabled - this would need one more CLI parameter
> * Populate schemas - yes by default, cannot be disabled
> 
> ```
>     parser.add_argument(
>         "--init-db",
>         action="store_true",
>         help="initializes the target database (runs the scripts ending with '.ddl' that don't start with '00_'"
>     )
> ```

From my side, I think this is the only thing that could be implemented, otherwise LGTM, I'm even gonna use it soon in Atum Service.

lsulak:
Not relevant anymore.

PR #70, opened 2023-10-11
Basic code structures and classes

benedeki:
These are just some very basic structures of the project

### DIFF
--- a/ultet/build.sbt
+++ b/ultet/build.sbt
@@ -1,0 +1,25 @@
+/*
+ * Copyright 2022 ABSA Group Limited
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


ultet/src/main/scala/.gitkeep
```diff
@@ -1 +0,0 @@
-To remove when actual content is present
```
Zejnilovic, line 1: what is this file?

ultet/build.sbt
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License.
+ */
+
+import Dependencies._
+import com.github.sbt.jacoco.report.JacocoReportSettings
+
+ThisBuild / organization := "za.co.absa.fa-db"
+
+
+lazy val scala212 = "2.12.17"
```
Zejnilovic, line 23: I would go for 2.13 since it is standalone tool, if you guys don't mind

benedeki: Not sure, if all eventual libs are available in 2.13
If you believe it won't be an issue, please, change in your PR on the main function.

ultet/src/main/scala/.gitkeep
benedeki: It was there, just to have the directory in git.

Zejnilovic (approved):

lsulak (approved):

dk1844 (approved):

PR #73, opened 2023-10-11
adding a TransactionGroup enum-like object

lsulak:


dk1844 (approved):

ultet/src/main/scala/za/co/absa/ultet/model/TransactionGroup.scala
```diff
@@ -0,0 +1,30 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.ultet.model
+
+
+sealed abstract class TransactionGroup
+
+object TransactionGroup {
```
benedeki, line 22: Use enumeration pleas, that offers order out of the box. We don't need extra properties anyway

ultet/src/main/scala/za/co/absa/ultet/model/TransactionGroup.scala
```diff
@@ -0,0 +1,30 @@
...
+  final case object Database      extends TransactionGroup
```
benedeki, line 23: Too many groups.
Enough is to have:
* Role
* Database
* Objects
* Indexes (might be possible to merge with Object maybe, but to be safe, let's keep separated).

lsulak: Hmm okay, when I think about it I agree - I wanted to be granular and explicit, but then sorting would be more complicated and not within one big transaction as planned for these objects

benedeki (approved):

PR #74, opened 2023-10-11
Feature/57 function commons

dk1844:
@lsulak @benedeki @jakipatryk


ultet/src/main/scala/za/co/absa/ultet/dbitems/DBFunction.scala
```diff
@@ -0,0 +1,27 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package za.co.absa.ultet.dbitems
+
+trait DBFunction extends DBItem {
+def fnName: String
+  def paramTypes: Seq[String]
```
jakipatryk, line 20: argument names will be needed too, and also output type(s);

I created something similar, perhaps we can merge them into one somewhere? https://github.com/AbsaOSS/fa-db/pull/75/files#diff-c3121d472cb3b94fccd4d0c299a444f9b596c5d81c77bed13aa1e1458fe1d4f5

ultet/src/main/scala/za/co/absa/ultet/dbitems/DBItemFromSource.scala
```diff
@@ -0,0 +1,39 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package za.co.absa.ultet.dbitems
+
+import za.co.absa.ultet.model.SQLEntry
+import za.co.absa.ultet.model.function.{FunctionBody, FunctionOwnership, FunctionGrant}
+
+case class DBItemFromSource(
```
jakipatryk, line 21: `DBFunctionFromSource`?

dk1844: Thanks, fixed

ultet/src/main/scala/za/co/absa/ultet/dbitems/DBFunction.scala
dk1844: Well, that depends - if you choose to recreate the SQL, then yes, otherwise, if you keep it as-is, you don't need it for signature comparison - only types are compared, I believe.

benedeki: Nope not needed. Just the types should be enough.

benedeki (approved):

PR #75, opened 2023-10-11
Fill in sqlExpression for functions related SQL entries

jakipatryk:


ultet/src/main/scala/za/co/absa/ultet/model/function/FunctionBody.scala
```diff
@@ -15,10 +15,45 @@
...
+       |  RETURNS $returnType
+       |AS
+       |$$$$
+       |BEGIN
+       |${bodyCode.value}
+       |END;
+       |$$$$
+       |LANGUAGE plpgsql VOLATILE SECURITY DEFINER;
+       |""".stripMargin
+  }
```
dk1844, line 54: I wonder if we need to recreate the function SQL at all. Why don't we just save it whole and pass it as-is?

dk1844 (approved):
LGTM

ultet/src/main/scala/za/co/absa/ultet/model/function/FunctionBody.scala
jakipatryk: Yeah maybe. But then we don't need drops, getting current function metadata from PG. Hmm

jakipatryk: I mean if we just do `CREATE OR REPLACE FUNCTION` in SQL file.

Zejnilovic (approved):

ultet/src/main/scala/za/co/absa/ultet/dbitems/DBFunctionFromPG.scala
```diff
@@ -15,18 +15,19 @@
  */
 package za.co.absa.ultet.dbitems
 
-import za.co.absa.ultet.model.SQLEntry
+import za.co.absa.ultet.model
+import za.co.absa.ultet.model.{SQLEntry, function}
 
 case class DBFunctionFromPG(sqlEntries: Seq[SQLEntry]) extends DBFunction {
-  override def fnName: String = ???
+  override def fnName: function.FunctionName = ???
```
benedeki, line 22: Why keep the prefixes?

ultet/src/main/scala/za/co/absa/ultet/model/function/FunctionGrant.scala
```diff
@@ -15,17 +15,25 @@
...
+       |  ${schemaName.value}.${functionName.value}($argumentsString)
+       |TO
+       |  ${userToGrantExecuteTo.value};
+       |""".stripMargin
+  }
 
   override def transactionGroup: TransactionGroup = ???
 
-  override def orderInTransaction: Int = ???
+  override def orderInTransaction: Int = 102
```
benedeki, line 38: **Minor:**
Would swap the order of grants and ownership

ultet/src/main/scala/za/co/absa/ultet/dbitems/DBFunctionFromPG.scala
jakipatryk: IntelliJ, fixed

PR #77, opened 2023-10-11
59 Add input parsers

### LINKED ISSUE #59: Input parameters processing & source files listing
## Background
The application has to be run, recognize input parameters and provide the list of files to process.

## Feature
Create the `main` function of the application. Including recognition of input parameters (consider what parameters might be needed) and  getting the list of file names to process.

It should then initialize the execution of the deployment process.

## Proposed Solution
Switches for:
1. Source files path
2. Dry run
3. Target DB+port+user+password
4. Connection string file to avoid the need to repeatedly enter the items of 3.
Secrets handling can probably be ignored for now.

Zejnilovic:
Closes #59


ultet/src/main/scala/za/co/absa/ultet/util/CliParser.scala
```diff
@@ -0,0 +1,51 @@
...
+        .required(),
+      opt[Boolean]('d', "dry-run")
+        .action((x, c) => c.copy(dryRun = x))
+        .text("Dry run option. Does not commit anything to DB.")
+        .optional(),
+      opt[String]('c', "db-connection-properties")
+        .action((x, c) => c.copy(dbConnectionPropertiesPath = x))
+        .text("Path to properties file with DB Connection info.")
+        .optional(),
+      opt[String]('p', "patch-file")
```
lsulak, line 45: btw, if this will run automatically in CICD, how / where these files will be saved? Potentially we can also add support to use S3 as the target? Just an idea, not for now but wanted to propose it

ultet/src/main/scala/za/co/absa/ultet/util/CliParser.scala
```diff
@@ -0,0 +1,51 @@
...
+      head("ultet", "0.1"),
+      opt[String]('s', "source-yaml")
+        .action((x, c) => c.copy(yamlSource = x))
+        .text("Path to source file(s). Use * as a wildcard.")
+        .required(),
+      opt[Boolean]('d', "dry-run")
+        .action((x, c) => c.copy(dryRun = x))
+        .text("Dry run option. Does not commit anything to DB.")
+        .optional(),
+      opt[String]('c', "db-connection-properties")
```
lsulak, line 41: Perhaps this can be sourced also from some AWS SEcret Manager or whatnot in the very distant future

lsulak (commented):
LGTM, I just read the code

ultet/src/main/scala/za/co/absa/ultet/util/CliParser.scala
```diff
@@ -0,0 +1,51 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.ultet.util
+
+import scopt.OParser
+
+case class Config(yamlSource: String = "",
```
lsulak, line 21: IntelliJ is complaining about this:

`Class doesn't correspond to file name `

perhaps it can be extracted (and renamed?) into its own file, e.g. `AppConfig`

Zejnilovic: Fixed

ultet/src/main/scala/za/co/absa/ultet/util/CliParser.scala
lsulak: What about having patch files stored somewhere even with the non-dry run though? Just to tract exactly what changed. This might be especially needed for the first few iterations of this tool - it might have bugs, unexpected scenarios, novel/beta features etc

ultet/src/main/scala/za/co/absa/ultet/util/DBProperties.scala
```diff
@@ -0,0 +1,74 @@
...
+    // Assume other properties are additional properties for the connection.
+    val propertiesConfig = config.getConfig("properties")
+    val properties = propertiesConfig.entrySet().map { entry =>
+      entry.getKey -> propertiesConfig.getString(entry.getKey)
+    }.toMap
+
+    (protocol, subprotocol) match {
+      case (Some(p), Some(s)) => DBProperties(serverName, database, portNumber, properties, p, s)
+      case (_, Some(s)) => DBProperties(serverName, database, portNumber, properties, subprotocol = s)
+      case (_,_) => DBProperties(serverName, database, portNumber, properties)
```
dk1844, line 84: I think this has the effect subprotocol is given, but protocol is not, subprotocol is silently ignored. Okayish for hackathon, I guess.

dk1844 (approved):
LGTM (just read the code)

PR #79, opened 2023-10-11
#64 class representing schema user

Zejnilovic:
Closes #64

### DIFF
--- a/ultet/src/main/scala/za/co/absa/ultet/model/schema/SchemaCreate.scala
+++ b/ultet/src/main/scala/za/co/absa/ultet/model/schema/SchemaCreate.scala
@@ -20,7 +20,7 @@ import za.co.absa.ultet.model.{SQLEntry, SchemaName, TransactionGroup}
 import za.co.absa.ultet.model.TransactionGroup.TransactionGroup
 
 case class SchemaCreate(name: SchemaName) extends SQLEntry {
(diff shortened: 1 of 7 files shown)


ultet/src/main/scala/za/co/absa/ultet/model/schema/SchemaCreate.scala
```diff
@@ -0,0 +1,28 @@
...
+ * limitations under the License.
+ */
+
+package za.co.absa.ultet.model.schema
+
+import za.co.absa.ultet.model.{SQLEntry, SchemaName, TransactionGroup}
+import za.co.absa.ultet.model.TransactionGroup.TransactionGroup
+
+case class SchemaCreate(name: SchemaName) extends SQLEntry {
+  override def sqlExpression: String = s"CREATE SCHEMA $name;"
```
jakipatryk, line 23:
```suggestion
  override def sqlExpression: String = s"CREATE SCHEMA ${name.value};"
```

dk1844 (approved):
LGTM

jakipatryk (approved):

benedeki (approved):
Merge it, but we need to eventually add a sequence of `DBItem` that belong to the schema.
(Or create a parent class joining `DBFunction` and `DBTableDiff`)

benedeki:
Also special handling of some schemas, like _public_

PR #84, opened 2023-10-11
65 run seq of sql entries

### LINKED ISSUE #65: Executing a sequence of SQL entries on the database
## Background
Once the analysis is done, it is to be executed on the database.

## Feature
A function that executes a sequence of SQL entries provided by the other application DB objects.
All the entries within the sequence are executed within one transaction.

Zejnilovic:
Closes #65

### DIFF
--- a/ultet/src/main/scala/za/co/absa/ultet/Ultet.scala
+++ b/ultet/src/main/scala/za/co/absa/ultet/Ultet.scala
@@ -22,11 +22,11 @@ import za.co.absa.ultet.util.{CliParser, Config, DBProperties}
 
 import java.nio.file._
 import scala.collection.JavaConverters._
-import java.sql.{Connection, ResultSet}
+import java.sql.{Connection, DriverManager, ResultSet}
 import scala.util.{Failure, Success, Try}
 
 object Ultet {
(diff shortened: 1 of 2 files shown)


ultet/src/main/scala/za/co/absa/ultet/Ultet.scala
```diff
@@ -17,12 +17,41 @@
...
 object Ultet {
+  def runTransaction(connection: Connection, entries: Seq[SQLEntry]): Seq[ResultSet] = {
+    val autoCommitOriginalStatus = connection.getAutoCommit
+    connection.setAutoCommit(false)
+
+    val resultSets = Try {
+      entries.foldLeft(List[ResultSet]()) { case (acc, entry) =>
+        val statement = connection.createStatement()
+        statement.execute(entry.sqlExpression)
+        val ret: List[ResultSet] = acc :+ statement.getResultSet
```
jakipatryk, line 37: very minor in our case, but `statement.getResultSet :: acc` and at the end reversing the list should yield better algo complexity (hopefully my brain isn't dead already today)

jakipatryk (approved):
Just checked code.

PR #85, opened 2023-10-11
66 ordering of sql queries

### LINKED ISSUE #66: Ordering of SQL entries
## Background
The order of the SQL entries is not automatic and have its own logics

## Feature
The full list of all SQL entries provided by all the objects needs to be ordered by their types (e.g. functions first) and grouped by transactions to be executed in.

## Proposed Solution
Returns a Sequence of objects. Each object represents a database (contains connection info) and transaction.

Zejnilovic:
Closes #66

### DIFF
--- a/ultet/src/main/scala/za/co/absa/ultet/Ultet.scala
+++ b/ultet/src/main/scala/za/co/absa/ultet/Ultet.scala
@@ -52,6 +52,13 @@ object Ultet {
     resultSets
   }
 
+  def sortEntries(entries: Seq[SQLEntry]): Seq[SQLEntry] = {
+    entries.sortWith {
+      case (a, b) if a.transactionGroup == b.transactionGroup => a.orderInTransaction < b.orderInTransaction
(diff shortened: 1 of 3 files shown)


ultet/src/main/scala/za/co/absa/ultet/Ultet.scala
```diff
@@ -26,6 +27,23 @@ import java.sql.{Connection, DriverManager, ResultSet}
 import scala.util.{Failure, Success, Try}
 
 object Ultet {
+  private val logger = Logger(getClass.getName)
+
+  private def runEntries(dbProperties: DBProperties, dbConnection: String, entries: Seq[SQLEntry]): Unit = {
+    val connection: Connection = DriverManager.getConnection(dbConnection, dbProperties.user, dbProperties.password)
```
jakipatryk, line 33: Would make it implicit at `Ultet` object level and set `autocommit` to true (I don't know what it is set to by default).

jakipatryk (approved):

PR #89, opened 2023-10-11
Implement basic DBItems creation from files URIs

### LINKED ISSUE #82: Creation of the model of DBItems
## Background
Once the source files has been identified they need to be changed to a tree of the model representing all kind of `DBItems`

## Feature
Provided the list of files and DB connection build up the model, resulting in a tree of `DBItems, returning the tree root.

It's up to your consideration what will be the function signature. If to accept a list of strings as file names. Or to get another structure like a `Map[String, Seq[String]` + `rootPath` where the keys of the map are the schemas and sequences the files within.

Also up to your consideration if related 'DBItem's are created recursively or explicitly by this function.
E,g, the `DBUserItem` which is the owner of a function will be returned as part of the `DBFunctionFromSource` creation.
Or explicitly called in this function because it known a function has an owner and list of users. (Probably the slightly preferred option is the recursive approach).

## Proposed Solution
Sets for collections seem to be an effective way to do, to ensure same items are not there multiple times.

NB!
Database creation has to be done on a different connection than all the other steps. But perhaps we can skip the DB creation at the moment or do it explicitly before calling this function.

NB!
Issue of schema merging. (Let's discuss when we get to it)

Depends on #

jakipatryk:
Closes #82

### DIFF
--- a/ultet/project/Dependencies.scala
+++ b/ultet/project/Dependencies.scala
@@ -21,6 +21,7 @@ object Dependencies {
     "org.scalactic" %% "scalactic" % "3.2.17",
     "org.scalatest" %% "scalatest" % "3.2.17" % "test",
     "com.github.scopt" %% "scopt" % "4.1.0",
-    "com.typesafe" % "config" % "1.4.2"
+    "com.typesafe" % "config" % "1.4.2",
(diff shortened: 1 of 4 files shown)


dk1844 (approved):
LGTM

PR #95, opened 2023-10-12
Do some things to glue together parts of the tool and make it working at least up to mocked runEntries

jakipatryk:


lsulak (approved):
LGTM

PR #96, opened 2023-10-12
#62 Creating table class from Postgres system tables

### LINKED ISSUE #62: Creating table class from Postgres system files
## Background
Tables need to be represented within the application for further analysis and eventually change SQL entries

## Feature
The table class is to be filled from Postgres system tables.

## Hint
This is for functions, but query or table won't be that different.
```sql
SELECT
    n.nspname,
	A.rolname,
	T.typname AS returntype,
    p.*
    
FROM 
    pg_catalog.pg_namespace n JOIN 
    pg_catalog.pg_proc p ON p.pronamespace = n.oid JOIN
	pg_catalog.pg_authid A ON  p.proowner  = A.oid JOIN
	pg_catalog.pg_type T ON p.prorettype = T.oid
WHERE 	
	n.nspname NOT IN ('public', 'pg_catalog', 'pg_toast', 'information_schema') AND
	p.prokind = 'f'
ORDER BY n.nspname, p.proname
```

Depends on #62

benedeki:
Closes #62


ultet/src/main/scala/za/co/absa/ultet/dbitems/DBTable.scala
```diff
@@ -40,3 +43,24 @@ case class DBTable(
     this.copy(primaryKey = Some(pk))
   }
 }
+
+object DBTable {
+  def createFromPG(schemaName: SchemaName, tableName: TableName, databaseName: DatabaseName)
+                  (implicit jdbcConnection: Option[Connection]): Option[DBTable] = {
```
jakipatryk, line 49: why Option?

PR #106, opened 2023-12-08
Feature/doobie with status

### LINKED ISSUE #11: Add Doobie module
## Background
Different DB libraries can be used to run the project

## Feature
Add a module of [Doobie ](https://github.com/tpolecat/doobie )executor engine.

### LINKED ISSUE #51: Simplify usage by offering combined classes

### LINKED ISSUE #99: Dependency propagation in DBSchema class
## Description of the bug

Requiring a `DBEngine` in the `DBSchema` class when it's not directly used can lead to unnecessary coupling and confusion. It can make the code harder to understand and maintain, as developers might spend time looking for where and how `DBEngine` is used within `DBSchema`. It also makes `DBSchema` harder to test, as any tests would need to provide a `DBEngine`, even though it's not used. If `DBEngine` is only needed for `DBFunction` instances, it would be better to require it where it's actually used. This would make the code easier to understand, maintain, and test. Indeed, not only is the `DBEngine` challenging to test due to its current implementation, but the code that utilizes `DBFunction` also faces similar testing difficulties. This is largely due to the propagated dependencies, which add unnecessary complexity to the testing process.

Passing dependencies through classes that don't use them, also known as **dependency propagation**, is generally considered a bad practice for several reasons:

1. **Violation of Single Responsibility Principle**: Each class should have a single responsibility. When a class takes a dependency only to pass it on to other classes, it's doing more than it should.

2. **Increased Coupling**: Dependency propagation increases the coupling between classes. Changes in one class can affect others, making the codebase more difficult to maintain.

3. **Harder to Understand**: It can be confusing for developers to see a class require a dependency that it doesn't use. They might spend unnecessary time trying to understand why that dependency is there.

4. **Difficult to Test**: Classes with unnecessary dependencies can be harder to test. You would need to mock more objects for unit tests, even if they are not used by the class under test.

This design choice of propagation of the `DBEngine` dependency through the `DBSchema` class, has inadvertently led to some challenges in several repositories that utilize this code (atum, aul). The added complexity and coupling, while not intentional, have made the code a bit more difficult to maintain and understand. A potential refactor could enhance the overall code quality and make it more efficient for future use. The coupling of dependencies in the `Runs` class (typical scenario in various existing codebases), as shown below, is considered a bad practice for several reasons:

1. **Testability**: Tightly coupled dependencies make unit testing more difficult. You can't easily replace `WriteCheckpoint` with a mock or stub for testing, because it's instantiated directly within the `Runs` class.

2. **Flexibility**: The `Runs` class is directly dependent on the `WriteCheckpoint` class. If you want to use a different implementation of the same functionality, you would need to modify the `Runs` class itself.

3. **Readability**: It's harder to understand the dependencies of the `Runs` class at a glance, because they're not explicitly listed as constructor parameters. You have to look inside the class to see that it depends on `WriteCheckpoint`.

4. **Reuse**: The `Runs` class can't be used with a different `WriteCheckpoint` instance. If you have another `WriteCheckpoint` instance with different settings or state, you can't use it with an existing `Runs` instance.

```scala
class Runs (implicit dBEngine: SlickPgEngine) extends DBSchema {
  import Runs._

  val writeCheckpoint = new WriteCheckpoint
  // other functions
  // ...
}
```

I would like to kindly propose the following fix. The `DBEngine` dependency, which is currently passed through the `DBSchema` class, could be removed from `DBSchema` and instead be passed directly to the `DBFunction` instances that require it. This would make the code easier to understand and maintain, as each class would only have the dependencies it directly uses. It would also make the code easier to test, as tests for `DBSchema` would no longer need to provide a `DBEngine`. This change would involve modifying the constructor `DBSchema`, and any code that creates instances of these classes.

```scala
abstract class DBSchema(schemaNameOverride: Option[String] = None)(implicit val namingConvention: NamingConvention)
{

  def this(schemaNameOverride: String)(implicit namingConvention: NamingConvention) {
    this(Option(schemaNameOverride))(namingConvention)
  }

  def this()(implicit namingConvention: NamingConvention) {
    this(None)(namingConvention)
  }

  def this(namingConvention: NamingConvention, schemaNameOverride: String) {
    this(Option(schemaNameOverride))(namingConvention)
  }

  // rest of the body
  // ...
}
```

With the proposed fix in place, database access could be modeled as follows: a `MoviesRepository` trait defines the database operations, and a `MoviesRepositoryImpl` class provides an implementation of this trait, using a `GetMovieById` instance to perform the actual database query.

```scala
trait MoviesRepository {
  def getMovieById(id: Int): IO[DatabaseError, Option[Movie]]
}

/**
 *  An implementation of the MoviesRepository trait.
 */
class MoviesRepositoryImpl(getMovieByIdFn: GetMovieById) extends MoviesRepository {

  /**
   *  Gets a movie by ID.
   */
  override def getMovieById(id: Int): IO[DatabaseError, Option[Movie]] = {
    ZIO
      .fromFuture { implicit ec: ExecutionContext => getMovieByIdFn(id) }
      .tap {
        case Some(movie) => ZIO.logInfo(s"Retrieved movie with ID $id: $movie")
        case None        => ZIO.logInfo(s"Movie with ID $id not found.")
      }
      .mapError(error => DatabaseError(error.getMessage))
      .tapError(error => ZIO.logError(s"Failed to retrieve movie with ID $id: ${error.message}"))
  }
}
```

```scala
class GetMovieById(implicit override val schema: DBSchema, val dbEngine: SlickPgEngine)
  extends DBOptionalResultFunction[Int, Movie, SlickPgEngine] // implicitly passed schema and dbEngine here
    with SlickFunction[Int, Movie] {

  // implementation
}
```
With this structure, `GetMovieById` can be easily mocked in tests, allowing for more controlled and isolated testing of the `MoviesRepositoryImpl` functionality. The `DBSchema` could be modelled as a singleton object that could be reused throughout application code if needed in multiple places.
```scala
object MoviesSchema extends DBSchema("schema name override")
// or object Runs extends DBSchema ... if we want to derive the name of the schema from the object name
```

With the suggested implementation, instance of `MoviesRepositoryImpl` will only have access to the `GetMovieById` function that it really needs. The **principle of least privilege** here will reduce the potential for errors or security breaches.

I firmly believe that implementing these changes will significantly enhance the quality of our code, making it more robust, maintainable, and testable.

### LINKED ISSUE #104: Support for 2.11 replaced by support for 2.13
Supporting older versions of Scala, in this case 2.11, can often be more of a burden than a benefit for several reasons:

1. **Deprecated Features and Libraries**: Scala 2.11 may contain features or libraries that have been deprecated or removed in newer versions. This can lead to code that is harder for us to maintain and upgrade in the future.

2. **Security Vulnerabilities**: Scala 2.11 and its libraries are more likely to contain security vulnerabilities. These vulnerabilities can be exploited by malicious actors to compromise the security of our applications. By using newer versions of Scala, we can benefit from the latest security patches and improvements.

3. **Performance Improvements**: Newer versions of Scala often come with performance improvements and optimizations. By not upgrading, we're missing out on these enhancements.

4. **Community Support**: The Scala community, like many other programming communities, tends to focus its support on the latest versions of the language. This means that we'll find more resources, such as tutorials, libraries, and tools, for newer versions of Scala.

5. **Future-Proofing**: By focusing on newer versions of Scala, such as 2.13 and Scala 3, we're future-proofing our code. We'll be able to take advantage of new features and improvements as soon as they're released.

In terms of security, using libraries full of vulnerabilities is a significant risk. Vulnerabilities can lead to a variety of issues, from data leaks to system compromises. It's crucial for us to keep our libraries up-to-date and replace those that are no longer maintained or have known vulnerabilities.

Supporting Scala 2.13 and potentially Scala 3 instead of 2.11 would be a more beneficial approach. Scala 2.13 has many improvements over 2.11, including better performance, new features, and fewer bugs. Scala 3, on the other hand, is a major release that introduces many changes and improvements, such as new syntax, better performance, and improved tooling. By supporting these versions, we'll be able to write more efficient, secure, and maintainable code.

### LINKED ISSUE #105: Modelling exceptions with Algebraic data types and handling status as Either[StatusException, R]
The current implementation of `checkStatus` results in `Try[FunctionStatus]`. The `Try` type is a wrapper for a computation that might fail or succeed. If the computation is successful, `Try` returns a `Success` containing the value, but if it fails, `Try` returns a `Failure` containing the exception.

The problem with this approach is that when a `Failure` is returned, it contains a `Throwable`, which is a very general type that can represent any kind of exception or error. This makes it difficult for client code to handle different kinds of errors in a specific way, because it can't pattern match on the resulting exception.

Furthermore, the type signature of `checkStatus` doesn't make it clear what kind of errors can occur. It just says that it might fail with some `Throwable`, but it doesn't specify what kind of `Throwable`. This lack of specificity can make it harder to understand how the function behaves just by looking at its type signature, and it can lead to errors being handled inappropriately or not being handled at all.

It's considered good practice in functional programming to make the error type explicit in the type signature of a function. This makes it clear what kinds of errors can occur, and it allows client code to handle different kinds of errors in a type-safe way. It also makes the function easier to understand and reason about, because its behavior is more explicitly described by its type signature.

I would like to propose modelling errors as ADTs and implement status checking that returns Either[StatusException, R] where R represents the final result type of the db function. 

The current exception model could be replace with:

```scala
sealed abstract class StatusException(val status: FunctionStatus) extends Exception(status.statusText)

final case class ServerMisconfigurationException(override val status: FunctionStatus) extends StatusException(status)
final case class DataConflictException(override val status: FunctionStatus) extends StatusException(status)
final case class DataNotFoundException(override val status: FunctionStatus) extends StatusException(status)
final case class ErrorInDataException(override val status: FunctionStatus) extends StatusException(status)
final case class OtherStatusException(override val status: FunctionStatus) extends StatusException(status)
```

The approach with case classes and a sealed abstract class has several advantages over using normal classes and extending from a common exception class like `DBFailException`:

1. **Pattern Matching**: Case classes are designed to be used with pattern matching. This allows you to destructure instances of case classes in a match expression and handle each case differently. This is a powerful feature that can make your code more readable and easier to understand.

2. **Immutability**: Case classes are immutable by default. This means that once a case class instance is created, it cannot be changed. This can help prevent bugs that are caused by mutable state.

3. **Equality**: Case classes automatically provide an `equals` method that compares instances by structure rather than by reference. This means that two instances of a case class with the same values are considered equal, even if they are different objects.

4. **Exhaustiveness Checking**: When you use a sealed abstract class, the compiler knows all possible subclasses of the class. This allows the compiler to check that all cases are covered in a match expression, which can help prevent bugs.

5. **Less Boilerplate**: Case classes automatically provide methods like `equals`, `hashCode`, and `toString`, as well as a companion object with `apply` and `unapply` methods. This can save you from writing a lot of boilerplate code.

Changing the error handling (status handling support) to result in `Either[StatusException, R]` would require a significant amount of refactoring. It's important to approach such changes incrementally to ensure that each step is correct and doesn't introduce new issues. We could start by first implementing the new `StatusException` hierarchy and then sketch out a plan of further refactoring.

salamonpavel:
In this PR, the below listed issues have been addressed:

- Issue #11
- Issue #51
- Issue #99
- Issue #104 
- Issue #105

I am aware that this pull request contains a substantial number of changes. I understand that this can make the review process more challenging and time-consuming, and I appreciate your efforts in ensuring our codebase maintains its high quality.

I am fully aware that the volume of changes may be overwhelming, and it's important to me that everyone on the team feels comfortable with these updates. To that end, I am more than willing to walk anyone through the code changes, explain the reasoning behind each modification, and answer any questions you may have. 

Please feel free to reach out to me at your convenience if you would like a walkthrough of the code or if you have any questions or concerns. 

Please review the changes and provide your feedback.

Closes #11 
Closes #51 
Closes #99 
Closes #104 
Closes #105


slick/src/main/scala/za/co/absa/fadb/slick/SlickQuery.scala
```diff
@@ -16,14 +16,64 @@
...
+/**
+ *  SQL query representation for Slick with status
+ *  @param sql        - the SQL query in Slick format
+ *  @param getResult  - the converting function, that converts the [[slick.jdbc.PositionedResult slick.PositionedResult]] (the result of Slick
+ *                   execution) into the desire `R` type
+ *  @tparam R         - the return type of the query
+ */
+class SlickQueryWithStatus[R](
+  val sql: SQLActionBuilder,
+  val getResult: GetResult[R],
```
salamonpavel, line 42: Probably not needed.

core/src/main/scala/za/co/absa/fadb/DBEngine.scala
```diff
@@ -16,57 +16,69 @@
...
   /**
-    * A type representing the (SQL) query within the engine
-    * @tparam T - the return type of the query
-    */
-  type QueryType[T] <: Query[T]
+   *  A type representing the (SQL) query within the engine
+   *  @tparam R - the return type of the query
+   */
+  type QueryType[R] <: Query[R]
+  type QueryWithStatusType[R] <: QueryWithStatus[_, _, R]
```
lsulak, line 37: perhaps these two can even live in `Query.scala`

core/src/main/scala/za/co/absa/fadb/status/handling/implementations/UserDefinedStatusHandling.scala
```diff
@@ -1,3 +1,5 @@
+package za.co.absa.fadb.status.handling.implementations
```
lsulak, line 1: I think that the licence should go first

core/src/test/scala/za/co/absa/fadb/status/handling/implementations/StandardStatusHandlingTest.scala
```diff
@@ -0,0 +1,99 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.fadb.status.handling.implementations
+
+import org.scalatest.funsuite.AnyFunSuiteLike
+import org.scalatest.matchers.should.Matchers.{a, convertToAnyShouldWrapper}
```
lsulak, line 20: I didn't know about this `a` but I don't think it's used

doobie/src/it/scala/za/co/absa/fadb/doobiedb/DoobieTest.scala
```diff
@@ -0,0 +1,46 @@
...
+  object Runs extends DBSchema
+
+  val printSqlLogHandler: LogHandler[IO] = new LogHandler[IO] {
+    def run(logEvent: LogEvent): IO[Unit] =
+      IO {
+        println(logEvent.sql)
+      }
+  }
+
+  protected val transactor = Transactor.fromDriverManager[IO](
```
lsulak, line 39: Type annotation required for protected member

doobie/src/it/scala/za/co/absa/fadb/doobiedb/DoobieTest.scala
```diff
@@ -0,0 +1,46 @@
...
+
+  import za.co.absa.fadb.naming.implementations.SnakeCaseNaming.Implicits._
+  object Runs extends DBSchema
+
+  val printSqlLogHandler: LogHandler[IO] = new LogHandler[IO] {
+    def run(logEvent: LogEvent): IO[Unit] =
+      IO {
+        println(logEvent.sql)
+      }
+  }
```
lsulak, line 37:
```suggestion
  val printSqlLogHandler: LogHandler[IO] = (logEvent: LogEvent) => IO {
    println(logEvent.sql)
  }
```

❓

doobie/src/main/scala/za/co/absa/fadb/doobiedb/DoobieEngine.scala
```diff
@@ -0,0 +1,87 @@
...
+   *  Executes a Doobie query and returns the result as an `F[Either[StatusException, R]]`.
+   *
+   *  @param query the Doobie query to execute
+   *  @param readStatusWithDataR the `Read[StatusWithData[R]]` instance used to read the query result into `StatusWithData[R]`
+   *  @return the query result as an `F[Either[StatusException, R]]`
+   */
+  private def executeQueryWithStatus[R](
+    query: QueryWithStatusType[R]
+  )(implicit readStatusWithDataR: Read[StatusWithData[R]]): F[Either[StatusException, R]] = {
+    query.fragment.query[StatusWithData[R]].unique.transact(transactor).map(query.getResultOrException)
```
lsulak, line 66: perhaps make a brief comment about this `unique` as we talked about, it raises an exception if the result does not have exactly 1 row - really bad naming if you ask me

doobie/src/main/scala/za/co/absa/fadb/doobiedb/StatusWithData.scala
```diff
@@ -0,0 +1,22 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package za.co.absa.fadb.doobiedb
+
+/**
+ *  Represents a function status with data.
+ */
+case class StatusWithData[R](status: Int, status_text: String, data: R)
```
lsulak, line 22:
```suggestion
case class StatusWithData[R](status: Int, statusText: String, data: R)
```

❓

doobie/src/main/scala/za/co/absa/fadb/doobiedb/package.scala
```diff
@@ -0,0 +1,42 @@
...
+import doobie.util.meta.Meta
+import doobie.implicits.javasql._
+
+import java.sql.Timestamp
+import java.time.ZoneId
+
+// can't be named doobie due to a naming conflict
+package object doobiedb {
+
+  object postgres {
```
lsulak, line 28: please verify that this is only related to PG. If not, then this library is even more universal :)

slick/src/main/scala/za/co/absa/fadb/slick/SlickFunction.scala
```diff
@@ -16,64 +16,96 @@
...
+   *  @return the `SlickQuery[R]` representing the SQL query
+   */
+  protected def query(values: I): SlickQuery[R] = new SlickQuery(sql(values), slickConverter)
+}
+
+private[slick] trait SlickFunctionWithStatus[I, R] extends SlickFunctionBase[I, R] {
+  /**
+   * Generates a `SlickQueryWithStatus[R]` representing the SQL query for the function with status support.
+   *
+   * @param status - the status to check
```
lsulak, line 65: not true right?

slick/src/main/scala/za/co/absa/fadb/slick/SlickFunction.scala
```diff
@@ -16,64 +16,96 @@
...
+  def checkStatus[A](statusWithData: FunctionStatusWithData[A]): Either[StatusException, A]
+}
+
+object SlickFunction {
+  /**
+   *  Class for Slick DB functions with status support.
+   */
+  abstract class SlickSingleResultFunctionWithStatus[I, R](functionNameOverride: Option[String] = None)(implicit
+    override val schema: DBSchema,
+    DBEngine: SlickPgEngine
```
lsulak, line 81:
```suggestion
    dBEngine: SlickPgEngine
```

also elsewhere, just to keep the convention

slick/src/main/scala/za/co/absa/fadb/slick/SlickPgEngine.scala
```diff
@@ -16,37 +16,49 @@
...
 
-
 import za.co.absa.fadb.DBEngine
 
 import scala.concurrent.{ExecutionContext, Future}
 import slick.jdbc.PostgresProfile.api._
+import cats.implicits._
 
 import scala.language.higherKinds
+import slick.jdbc.{GetResult, PositionedResult}
```
lsulak, line 26: those two seem not be needed

slick/src/main/scala/za/co/absa/fadb/slick/SlickPgEngine.scala
```diff
@@ -16,37 +16,49 @@
...
+ */
+class SlickPgEngine(val db: Database)(implicit val executor: ExecutionContext) extends DBEngine[Future] {
 
   /**
-    * The type of Queries for Slick
-    * @tparam T - the return type of the query
-    */
-  type QueryType[T] = SlickQuery[T]
+   *  The type of Queries for Slick
+   *  @tparam T - the return type of the query
```
lsulak, line 38:
```suggestion
   *  @tparam R - the return type of the query
```

slick/src/main/scala/za/co/absa/fadb/slick/SlickQuery.scala
```diff
@@ -16,14 +16,64 @@
...
 /**
-  * SQL query representation for Slick
-  * @param sql        - the SQL query in Slick format
-  * @param getResult  - the converting function, that converts the [[slick.jdbc.PositionedResult slick.PositionedResult]] (the result of Slick
-  *                   execution) into the desire `R` type
-  * @tparam R         - the return type of the query
-  */
+ *  SQL query representation for Slick
+ *  @param sql        - the SQL query in Slick format
+ *  @param getResult  - the converting function, that converts the [[slick.jdbc.PositionedResult slick.PositionedResult]] (the result of Slick
```
lsulak, line 27: quite long line (below as well)

slick/src/main/scala/za/co/absa/fadb/slick/SlickQuery.scala
```diff
@@ -16,14 +16,64 @@
...
+   * @param initialResult - the initial result of the query
+   * @return the status with data
+   */
+  override def processStatus(initialResult: PositionedResult): FunctionStatusWithData[PositionedResult] = {
+    val status: Int = initialResult.<<
+    val statusText: String = initialResult.<<
+    FunctionStatusWithData(FunctionStatus(status, statusText), initialResult)
+  }
+
+  /*
```
lsulak, line 57:
```suggestion
  /**
```

? also on other places, just to have it more colorful and consistent in IntelliJ

lsulak (commented):
All right review finished. I'll build & ran tests tomorrow. I spent a majority of today with it but I don't think I dig deep enough to provide really thought-provoking feedback. 

I generally speaking really like it, the approach with `Either` is quite interesting, among many other things. I especially like testability of this solution and that you removed some 'boilerplate' stuff.

--

I found a few minor things, I see that the Licence Check is failing, and also there is something with Jacoco CI run.

But once these are addressed, I think I'll approve it (I'll give it another iteration afterwards). I'm quite interested in @benedeki opinion.

.scalafmt.conf
```diff
@@ -0,0 +1,32 @@
...
+
+lineEndings = unix
+
+docstrings.blankFirstLine = yes
+docstrings.style = AsteriskSpace
+docstrings.wrap = no
+docstrings.removeEmpty = true
+
+align.openParenDefnSite = false
+align.openParenCallSite = false
```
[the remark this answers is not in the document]
salamonpavel: newline added

doobie/src/it/database/other_types.sql
```diff
@@ -0,0 +1,100 @@
...
+    status := 11;
+    status_text := 'ok';
+    o_id := p_id;
+  EXCEPTION WHEN unique_violation THEN
+    status := 31;
+    status_text := 'data conflict';
+  END;
+  RETURN NEXT;
+END;
+$$ LANGUAGE plpgsql;
```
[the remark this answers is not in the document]
salamonpavel: newline added

doobie/src/it/database/get_actors.sql
```diff
@@ -0,0 +1,36 @@
...
+RETURN QUERY SELECT A.actor_id, A.first_name, A.last_name
+                 FROM runs.actors A
+                 WHERE
+                     (i_first_name IS NULL OR A.first_name = i_first_name)
+                   AND
+                     (i_last_name IS NULL OR A.last_name = i_last_name)
+                 ORDER BY A.actor_id ASC;
+END;
+$$
+LANGUAGE plpgsql;
```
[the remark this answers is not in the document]
salamonpavel: added

doobie/src/it/database/get_actor_by_id.sql
```diff
@@ -0,0 +1,51 @@
...
+    last_name               VARCHAR(150)
+) AS
+$$
+BEGIN
+RETURN QUERY SELECT A.actor_id, A.first_name, A.last_name
+    FROM runs.actors A
+    WHERE A.actor_id = i_actor_id;
+END;
+$$
+LANGUAGE plpgsql;
```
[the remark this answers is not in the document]
salamonpavel: added

doobie/src/it/database/dates_times.sql
```diff
@@ -0,0 +1,122 @@
...
+                 p_sql_timestamp,
+                 p_util_date
+             ) RETURNING id INTO o_id;
+
+    status := 11;
+    status_text := 'OK';
+
+    RETURN;
+END;
+$$ LANGUAGE plpgsql;
```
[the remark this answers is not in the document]
salamonpavel: added

doobie/src/it/database/actors.ddl
```diff
@@ -0,0 +1,21 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+CREATE TABLE IF NOT EXISTS runs.actors(
+    actor_id INT GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
+    first_name VARCHAR(150) NOT NULL,
+    last_name VARCHAR(150) NOT NULL
+);
```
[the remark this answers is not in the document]
salamonpavel: added

doobie/src/it/database/create_actor.sql
```diff
@@ -0,0 +1,53 @@
...
+    VALUES (i_first_name, i_last_name) RETURNING actor_id
+    INTO o_actor_id;
+
+    status:= 11;
+    status_text:= 'Actor created';
+
+    RETURN;
+END;
+$$
+LANGUAGE plpgsql;
```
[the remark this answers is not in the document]
salamonpavel: added

core/src/main/scala/za/co/absa/fadb/DBEngine.scala
salamonpavel: I believe they have to be defined like this in DBEngine, otherwise we wouldn't be able to reference them as dBEngine.QueryType[R] and dBEngine.QueryWithStatusType[R] in DBFunction, respectively DBFunctionWithStatus.

core/src/main/scala/za/co/absa/fadb/DBFunction.scala
```diff
@@ -16,228 +16,189 @@
...
+  protected def query(values: I): dBEngine.QueryType[R]
+}
+
+/**
+ *  `DBFunctionWithStatus` is an abstract class that represents a database function with a status.
+ *  It extends the [[DBFunction]] class and adds handling for the status of the function invocation.
+ *  @param functionNameOverride - Optional parameter to override the class name if it does not match the database function name.
+ *  @param schema - The schema the function belongs to.
+ *  @param dBEngine - The database engine that is supposed to execute the function (contains connection to the database).
+ *  @param queryStatusHandling - The [[QueryStatusHandling]] instance that handles the status of the function invocation.
```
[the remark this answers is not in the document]
salamonpavel: good catch, removed it

core/src/main/scala/za/co/absa/fadb/status/handling/implementations/UserDefinedStatusHandling.scala
salamonpavel: thanks, I have put the license on top ...

core/src/test/scala/za/co/absa/fadb/status/handling/implementations/StandardStatusHandlingTest.scala
salamonpavel: correct, it's not used anymore in the test cases, removed it

doobie/src/it/scala/za/co/absa/fadb/doobiedb/DoobieTest.scala
salamonpavel: ok refactored in this fashion as one-liner

doobie/src/main/scala/za/co/absa/fadb/doobiedb/DoobieEngine.scala
salamonpavel: added comment and link to the docs

slick/src/main/scala/za/co/absa/fadb/slick/SlickFunction.scala
salamonpavel: thanks, fixed

slick/src/main/scala/za/co/absa/fadb/slick/SlickFunction.scala
salamonpavel: fixed

slick/src/main/scala/za/co/absa/fadb/slick/SlickPgEngine.scala
salamonpavel: indeed, not needed anymore

slick/src/main/scala/za/co/absa/fadb/slick/SlickQuery.scala
salamonpavel: ok changed both

slick/src/main/scala/za/co/absa/fadb/slick/SlickQuery.scala
salamonpavel: fixed

doobie/src/main/scala/za/co/absa/fadb/doobiedb/package.scala
salamonpavel: hopefully will get to it soon

doobie/src/it/scala/za/co/absa/fadb/doobiedb/DoobieTest.scala
salamonpavel: After downgrade of Doobie version this is no longer relevant. The support for LogHandler is not available in the downgraded version.

doobie/src/main/scala/za/co/absa/fadb/doobiedb/package.scala
salamonpavel: After downgrade of Doobie version this is no longer relevant. Package object was removed as it became obsolete.

.github/workflows/build.yml
```diff
@@ -29,14 +29,14 @@ jobs:
...
         include:
-          - scala: 2.11.12
-            scalaShort: "2.11"
-            overall: 0.0
-            changed: 80.0
           - scala: 2.12.17
             scalaShort: "2.12"
             overall: 0.0
             changed: 80.0
+          - scala: 2.13.11
```
jakipatryk, line 36: Why not 2.13.12?

core/src/main/scala/za/co/absa/fadb/DBEngine.scala
```diff
@@ -16,57 +16,69 @@
...
 
-  implicit val executor: ExecutionContext
+  /**
+   *  The actual query executioner of the queries of the engine
+   *  @param query  - the query to execute
+   *  @tparam R     - return the of the query
```
jakipatryk, line 42:
```suggestion
   *  @tparam R     - return type of the query
```

core/src/main/scala/za/co/absa/fadb/DBEngine.scala
```diff
@@ -16,57 +16,69 @@
...
   /**
-    * The actual query executioner of the queries of the engine
-    * @param query  - the query to execute
-    * @tparam R     - return the of the query
-    * @return       - sequence of the results of database query
-    */
-  protected def run[R](query: QueryType[R]): Future[Seq[R]]
+   *  The actual query executioner of the queries of the engine with status
+   *  @param query  - the query to execute
+   *  @tparam R     - return the of the query
```
jakipatryk, line 50:
```suggestion
   *  @tparam R     - return type of the query
```

core/src/main/scala/za/co/absa/fadb/DBEngine.scala
```diff
@@ -16,57 +16,69 @@
...
+   *  @return       - sequence of the results of database query
```
jakipatryk, line 51: So should it return a sequence or not? I believe now it returns a single result.

jakipatryk (commented):
I reviewed only part of the PR, but anyway I don't feel I know enough internals of fa-db to approve.

.github/workflows/build.yml
salamonpavel: sure, will use 2.13.12

core/src/main/scala/za/co/absa/fadb/DBEngine.scala
salamonpavel: good catch, it returns single result

core/src/main/scala/za/co/absa/fadb/DBEngine.scala
salamonpavel: `QueryType` is a type member of the `DBEngine` abstract class. It's a higher-kinded type, meaning it's a type that takes another type as a parameter. 

The `<: Query[R]` part is a type bound. It means that whatever type `QueryType[R]` is, it must be a subtype of `Query[R]`. `Query[R]` is presumably another type that represents a database query that returns results of type `R`.

In the context of `DBEngine`, `QueryType` is used to abstract over the specific type of queries that a particular database engine can execute. Different implementations of `DBEngine` can use different types of queries, as long as they are subtypes of `Query[R]`.

For example, one implementation of `DBEngine` might use SQL queries, while another might use NoSQL queries. These different types of queries can have different methods and properties, but as long as they are subtypes of `Query[R]`, they can be used with `DBEngine`.

This is a powerful feature of Scala's type system that allows for a high level of abstraction and flexibility in your code. One of the key features of type members is that they can be overriden in subclasses or subtraits.

lsulak:
File `/home/runner/work/fa-db/fa-db/examples/src/main/scala/za/co/absa/fadb/examples/enceladus/DatasetSchema.scala` still doesn't have licence header according to the CI

lsulak:
Jacoco doesn't seem to work here, can you take a look @miroslavpojer please? Perhaps this is something you encountered before

core/src/test/scala/za/co/absa/fadb/naming/ExplicitNamingRequiredTest.scala
```diff
@@ -0,0 +1,32 @@
+package za.co.absa.fadb.naming
```
lsulak, line 1: licence header missing

core/src/test/scala/za/co/absa/fadb/naming/implementations/AsIsNamingTest.scala
```diff
@@ -0,0 +1,20 @@
+package za.co.absa.fadb.naming.implementations
```
lsulak, line 1: licence header missing

core/src/test/scala/za/co/absa/fadb/naming/ExplicitNamingRequiredTest.scala
salamonpavel: thanks, added the license

core/src/test/scala/za/co/absa/fadb/naming/implementations/AsIsNamingTest.scala
salamonpavel: thanks, added the license

salamonpavel:
> File `/home/runner/work/fa-db/fa-db/examples/src/main/scala/za/co/absa/fadb/examples/enceladus/DatasetSchema.scala` still doesn't have licence header according to the CI

Seems that the license formatting was off. The license check passes now.

jakipatryk (approved):
LGTM (just read the code), though as I said I don't have much fa-db internals knowledge.

jakipatryk (approved):

PR #108, opened 2023-12-22
streaming functions for doobie and slick with tests

### LINKED ISSUE #107: Fetching database records with streaming approach [RESEARCH]
## Background
At the moment if we're fetching a large number of records from a database without using streaming, we can potentially run into an OutOfMemoryError. This is because all the records are loaded into memory at once.

## Feature
Streaming, on the other hand, loads data in chunks, keeping only a portion of the data in memory at any given time. This is more memory-efficient and can handle large datasets that would not fit into memory all at once.

Both Doobie and Slick support streaming. Moreover streaming integration with ZIO is possible. It would be great to explore the space and find out if/how we could introduce streaming capabilities into this library.

salamonpavel:
Added support for streaming for both Slick and Doobie module. Since Slick works with Futures but for streaming it relies on IOs or other effect types there was a need to abstract the streaming operation in a new DBStreamingEngine. 

Closes #107

### DIFF
--- a/core/src/main/scala/za/co/absa/fadb/DBFunction.scala
+++ b/core/src/main/scala/za/co/absa/fadb/DBFunction.scala
@@ -128,6 +128,28 @@ abstract class DBFunctionWithStatus[I, R, E <: DBEngine[F], F[_]](functionNameOv
   override def checkStatus[A](statusWithData: FunctionStatusWithData[A]): Either[StatusException, A]
 }
 
+abstract class DBStreamingFunction[I, R, E <: DBStreamingEngine[F], F[_]](functionNameOverride: Option[String] = None)(
+  implicit override val schema: DBSchema,
+  val dbStreamingEngine: E
+) extends DBFunctionFabric(functionNameOverride) {
+
+  // A constructor that takes only the mandatory parameters and uses default values for the optional ones
+  def this()(implicit schema: DBSchema, dBEngine: E) = this(None)
+
+  // A constructor that allows specifying the function name as a string, but not as an option
+  def this(functionName: String)(implicit schema: DBSchema, dBEngine: E) = this(Some(functionName))
+
+  protected def streamResults(values: I): fs2.Stream[F, R] = dbStreamingEngine.runStreaming(query(values))
+
+  /**
+   *  Function to create the DB function call specific to the provided [[DBEngine]].
+   *  Expected to be implemented by the DBEngine specific mix-in.
+   *  @param values - The values to pass over to the database function.
+   *  @return - The SQL query in the format specific to the provided [[DBEngine]].
+   */
+  protected def query(values: I): dbStreamingEngine.QueryType[R]
+}
+
 object DBFunction {
 
   /**
@@ -152,6 +174,20 @@ object DBFunction {
      *               type `R`
      */
     def apply(values: I): F[Seq[R]] = multipleResults(values)
+  }
+
+  abstract class DBStreamingResultFunction[I, R, E <: DBStreamingEngine[F], F[_]](
+    functionNameOverride: Option[String] = None
+  )(implicit schema: DBSchema, dBStreamingEngine: E)
+      extends DBStreamingFunction[I, R, E, F](functionNameOverride) {
+
+    // A constructor that takes only the mandatory parameters and uses default values for the optional ones
+    def this()(implicit schema: DBSchema, dBStreamingEngine: E) = this(None)
+
+    // A constructor that allows specifying the function name as a string, but not as an option
+    def this(functionName: String)(implicit schema: DBSchema, dBStreamingEngine: E) = this(Some(functionName))
+
+    def apply(values: I): fs2.Stream[F, R] = streamResults(values)
   }
 
   /**
--- a/project/Dependencies.scala
+++ b/project/Dependencies.scala
@@ -21,6 +21,7 @@ object Dependencies {
   private def commonDependencies(scalaVersion: String): Seq[ModuleID] = Seq(
     "org.typelevel" %% "cats-core" % "2.9.0",
     "org.typelevel" %% "cats-effect" % "3.5.0",
+    "co.fs2" %% "fs2-core" % "3.7.0",
     "org.scalatest"      %% "scalatest" % "3.1.0"           % "test,it",
     "org.scalatest"      %% "scalatest-flatspec" % "3.2.0"  % "test,it",
     "org.scalatestplus"  %% "mockito-1-10" % "3.1.0.0"      % "test,it"
@@ -39,7 +40,8 @@ object Dependencies {
       "org.slf4j"            % "slf4j-nop"                    % "1.7.26",
       "com.typesafe.slick"  %% "slick-hikaricp"               % "3.3.3",
       "org.postgresql"       % "postgresql"                   % "42.6.0",
-      "com.github.tminglei" %% "slick-pg"                     % "0.20.4"   % Optional
+      "com.github.tminglei" %% "slick-pg"                     % "0.20.4"   % Optional,
+      "co.fs2" %% "fs2-reactive-streams" % "3.9.3"
     )
   }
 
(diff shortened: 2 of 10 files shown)


core/src/main/scala/za/co/absa/fadb/DBFunction.scala
```diff
@@ -74,7 +74,7 @@ abstract class DBFunction[I, R, E <: DBEngine[F], F[_]](functionNameOverride: Op
 
 /**
  *  `DBFunctionWithStatus` is an abstract class that represents a database function with a status.
- *  It extends the [[DBFunction]] class and adds handling for the status of the function invocation.
+ *  It extends the `DBFunction` class and adds handling for the status of the function invocation.
```
lsulak, line 77: What's the rule behind this? I can see references to abstract class (in this example) with both, `[[` and ```, across the code

doobie/src/main/scala/za/co/absa/fadb/doobie/DoobieStreamingEngine.scala
```diff
@@ -0,0 +1,71 @@
...
+    extends DBStreamingEngine[F] {
+
+  /** The type of Doobie queries that produce `R` */
+  type QueryType[R] = DoobieQuery[R]
+
+  /**
+   *  Executes a Doobie query and returns the result as an `fs2.Stream[F, R]`.
+   *
+   *  @param query the Doobie query to execute
+   *  @param readR the `Read[R]` instance used to read the query result into `R`
```
lsulak, line 42: this is not a separate parameter, but if you want to keep it you can modify this (here and below) somehow so that it's all under `query` param docs

project/Dependencies.scala
```diff
@@ -39,7 +40,8 @@ object Dependencies {
       "org.slf4j"            % "slf4j-nop"                    % "1.7.26",
       "com.typesafe.slick"  %% "slick-hikaricp"               % "3.3.3",
       "org.postgresql"       % "postgresql"                   % "42.6.0",
-      "com.github.tminglei" %% "slick-pg"                     % "0.20.4"   % Optional
+      "com.github.tminglei" %% "slick-pg"                     % "0.20.4"   % Optional,
+      "co.fs2" %% "fs2-reactive-streams" % "3.9.3"
```
lsulak, line 44:
```suggestion
      "co.fs2"              %% "fs2-reactive-streams"         % "3.9.3"
```

slick/src/it/scala/za/co/absa/fadb/slick/SlickStreamingResultFunctionTest.scala
```diff
@@ -0,0 +1,49 @@
...
+import za.co.absa.fadb.slick.SlickFunction.SlickStreamingResultFunction
+import za.co.absa.fadb.slick.FaDbPostgresProfile.api._
+
+class SlickStreamingResultFunctionTest extends AnyFunSuite with SlickTest {
+
+  class GetActors(implicit override val schema: DBSchema, val dbEngine: SlickPgStreamingEngine[IO])
+      extends SlickStreamingResultFunction[GetActorsQueryParameters, Actor, IO]
+      with ActorSlickConverter {
+
+    override def fieldsToSelect: Seq[String] = super.fieldsToSelect ++ Seq("actor_id", "first_name", "last_name")
```
lsulak, line 33: I think scala (or, at least IntelliJ) is confused here between immutable and mutable Seq. You don't have this problem here?

lsulak (commented):
Found a few tiny things and Jacoco coverage action is still failing (not on insufficient coverage though), so please take a look. Otherwise LGTM.

I pulled it, reviewed, compiled and ran tests.

core/src/main/scala/za/co/absa/fadb/DBFunction.scala
salamonpavel: The square brackets are better since when you change class name it updates automatically, that's not the case with backtics. I have replaced the backtics for brackets everywhere for object names.

doobie/src/main/scala/za/co/absa/fadb/doobie/DoobieStreamingEngine.scala
salamonpavel: Thanks, didn't notice it. Removed it from methods where the read instance is not one of parameters.

project/Dependencies.scala
salamonpavel: Alignments in Dependencies.scala fixed.

project/Dependencies.scala
```diff
@@ -21,6 +21,7 @@ object Dependencies {
   private def commonDependencies(scalaVersion: String): Seq[ModuleID] = Seq(
     "org.typelevel" %% "cats-core" % "2.9.0",
     "org.typelevel" %% "cats-effect" % "3.5.0",
+    "co.fs2" %% "fs2-core" % "3.7.0",
```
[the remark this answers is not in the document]
salamonpavel: Fixed as mentioned above.

slick/src/it/scala/za/co/absa/fadb/slick/SlickStreamingResultFunctionTest.scala
salamonpavel: Yeah, it's probably just IntelliJ's compiler confused. The code compiles without any problem with scalac compiler.

salamonpavel:
> Found a few tiny things and Jacoco coverage action is still failing (not on insufficient coverage though), so please take a look. Otherwise LGTM.
> 
> I pulled it, reviewed, compiled and ran tests.

Fixed this. The problem was in node version Github's updated runner uses now. Solved by an update of the jacoco github action version.

salamonpavel:
@lsulak @jakipatryk Thanks for the initial review here and on the calls we had yesterday. Based on your comments I have made the streaming part for Slick as a separate package since additional dependencies are needed and we don't want to be importing transitive dependencies into our projects when streaming is not used. At the moment there is core module for non-streaming, streaming module for streaming, slick module for slick non-streaming, slick-streaming module for slick streaming and doobie module. Doobie module provides both non-streaming and streaming functions since Doobie already bring all the dependencis needed (works with fs2 streams under the hood; cats effect brought in transitively from fs2) so there would be no benefit in splitting this into separate modules.

.github/workflows/build.yml
```diff
@@ -93,7 +97,9 @@ jobs:
         with:
           paths: >
             ${{ github.workspace }}/core/target/scala-${{ matrix.scalaShort }}/jacoco/report/jacoco.xml,
+            ${{ github.workspace }}/streaming/target/scala-${{ matrix.scalaShort }}/jacoco/report/jacoco.xml,
             ${{ github.workspace }}/slick/target/scala-${{ matrix.scalaShort }}/jacoco/report/jacoco.xml
```
miroslavpojer, line 101:
```suggestion
            ${{ github.workspace }}/slick/target/scala-${{ matrix.scalaShort }}/jacoco/report/jacoco.xml,
```

.github/workflows/build.yml
```diff
@@ -93,7 +97,9 @@ jobs:
...
+            ${{ github.workspace }}/slick-streaming/target/scala-${{ matrix.scalaShort }}/jacoco/report/jacoco.xml
```
miroslavpojer, line 102:
```suggestion
            ${{ github.workspace }}/slick-streaming/target/scala-${{ matrix.scalaShort }}/jacoco/report/jacoco.xml,
```

miroslavpojer (commented):
- [ ] pulled
- [ ] built
- [ ] test run
- [ ] jacoco parts - code review

.github/workflows/build.yml
```diff
@@ -93,7 +97,9 @@ jobs:
...
             ${{ github.workspace }}/doobie/target/scala-${{ matrix.scalaShort }}/jacoco/report/jacoco.xml
```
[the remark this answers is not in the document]
salamonpavel: @miroslavpojer I am not satisfied with this situation either. Unfortunately I couldn't perform merge of previous pull request with this check in place. Let's work together to find a good solution so we can perform tests against a database both locally and within a Github pipeline; for instance with a dockerized Postgres instance.

lsulak (approved):
LGTM, just read the code

Issue #118, opened 2024-03-21
Provide `DoobieMultipleResultFunctionWithStatus`


lsulak:
## Background
See the description of this feature in https://github.com/AbsaOSS/atum-service/issues/169 where the idea emerged.

Issue #116, opened 2024-02-06
Update README files for Slick and Doobie INT tests


miroslavpojer:
## Background
The project contains integration tests for Slick and Doobie modules.
These tests need some preparation steps to be able to run.
These steps are not well documented.
I tried to get them running and I failed on missing troubleshooting chapter and on missing test data inserts.

## Feature
- Update both README files to provide a list of steps to get tests running.
- Provide the SQL file content with INSERT to the empty DB.

lsulak:
Closing as done in the #118 and if there is more need to add something more, we can always reopen this.

Issue #19, opened 2022-04-10
Database deployment capabilities


benedeki:
## Background
For effective usage of _fa-db_ it expects the database to have the actual functions and tables they use. Manual deployment is tedious and error prone. Automation would be very helpful.

## Goal
Ability to deploy or upgrade DB to the current version with automatic generation of structure transformation queries and functions deployment.

## Expected Task List [Optional]
A list of expected issues that will be needed to achieve this Epic
1. Ability to setup the database
2. Ability to deploy functions
3. Ability to deploy tables and schemas
4. Unifying function to call all the above in one simple step.

PR #119, opened 2024-03-25
#118: DB cleanup and IT tests addition to CI

lsulak:
Preparation for #118 - I don't really want to do all the DB setup for tests manually, it's too much effort. I used Flyway to automate it. It's not ideal - the actual data insertions and cleanups after each test - but it's better than it was and we can improve it on the next increment.

What was done:
* DB files (SQL and DDL) were moved and renamed / polished to be compatible with Flyway tool; it seems like a lot of changes and files, but really it was mostly small changes
* I created `.github/workflows/integration_tests.yml` - but there are many things to improve in the future:
  * test coverage
  * better documentation perhaps (#116)
  * better test isolation - test data init and cleanup (#121)

--

Relates to #118 
Relates to #116
Relates to #121
Relates to #19


salamonpavel (dismissed):
Good job, this was definitely missing. Let's address this also in other projects.

database/README.md
```diff
@@ -0,0 +1,41 @@
+## About
```
benedeki, line 1: **To consider:**
I would rename the module to `test_database` or `demo_database`.

build.sbt
```diff
@@ -48,6 +48,14 @@ lazy val commonJacocoReportSettings: JacocoReportSettings = JacocoReportSettings
 lazy val commonJacocoExcludes: Seq[String] = Seq(
 )
 
+enablePlugins(FlywayPlugin)
+flywayUrl := FlywayConfiguration.flywayUrl
+flywayUser := FlywayConfiguration.flywayUser
+flywayPassword := FlywayConfiguration.flywayPassword
+flywayLocations := FlywayConfiguration.flywayLocations
+flywaySqlMigrationSuffixes := FlywayConfiguration.flywaySqlMigrationSuffixes
+libraryDependencies ++= flywayDependencies
```
benedeki, line 57: I think it adds to the global dependencies, doesn't it? I think flyway dependencies shouldn't be a generic dependency.

build.sbt
```diff
@@ -48,6 +48,14 @@ lazy val commonJacocoReportSettings: JacocoReportSettings = JacocoReportSettings
...
+
```
benedeki, line 58: `database` module is not present. Also isn't it overlapping the modules `database` and `examples`?

database/README.md
lsulak: Okay, agreed, will do that

build.sbt
lsulak: I believe this is what @salamonpavel was dealing with in the past and there was no way how to circumvent this (I know :( ) by extracting / moving elsewhere. Anyway, it's part of the build, i.e. general dependency: https://github.com/AbsaOSS/atum-service/blob/master/build.sbt#L53-L59

(remember, Flyway is temporary anyway)

build.sbt
lsulak: good point, an overlook on my side about missing `database` - in terms of examples, I didn't care about them here in this PR, but why not, I'll think about deleting / merging or moving them.

lsulak: based on the content of those tests in examples/, I decided to drop them completely. All the tests there are ignored, it refers to `menas` DB, and objects that are not in this repo, and those individual functions that would be tested are basically nicely covered in actual tests in Slick.

benedeki: Good, no objections

benedeki (approved):
Well, except the Flyway dependency I think it's OK. That one can be addressed later.

build.sbt
Zejnilovic: I agree with @benedeki 

```sbt
lazy val modulNeedingFlyaway = project
  .enablePlugins(FlywayPlugin)
  .settings(
    flywayUrl := FlywayConfiguration.flywayUrl
    flywayUser := FlywayConfiguration.flywayUser
    flywayPassword := FlywayConfiguration.flywayPassword
    flywayLocations := FlywayConfiguration.flywayLocations
    flywaySqlMigrationSuffixes := FlywayConfiguration.flywaySqlMigrationSuffixes
    libraryDependencies ++= flywayDependencies
  )
```

and then run it like `sbt modulNeedingFlyaway/flywayMigrate`

benedeki: Can we redefine the `flywayMigrate` to behave like `sbt modulNeedingFlyaway/flywayMigrate`?

Zejnilovic: Like this `addCommandAlias("flywayMigrate", "modulNeedingFlyaway/flywayMigrate")`?

lsulak: Let me try, thanks @Zejnilovic. To be fair I haven't been playing with this one as my primary objective here was to actually have a IT suite embedded in our CI - but I agree that we can improve this mechanism in build.sbt

benedeki: That's why it got approved, not blocking the whole thing 😉

PR #120, opened 2024-03-25
#118: adding support for MultipleResultFunction with status for Doobie

### LINKED ISSUE #115: Add the ability to retrieve status code of a function on non-error return
## Background
Currently, from Scala, we only can retrieve resulting status code and status text of a given called function when there is an error, so we can work with error status codes, but we can't work with the success status codes - this might be important if we want to use it in Scala for some business logic or so.

### LINKED ISSUE #118: Provide `DoobieMultipleResultFunctionWithStatus`

lsulak:
What was done:
* In core, changed / added: `DBMultipleResultFunctionWithStatus`, `DBSingleResultFunctionWithStatus` and `DBOptionalResultFunctionWithStatus` based on `DBFunctionWithStatus` as it was for non-status classes, i.e. keeping the design consistent across the core module
* In Doobie added: `DoobieOptionalResultFunctionWithStatus` and `DoobieMultipleResultFunctionWithStatus`
* In Slick, added:  `SlickOptionalResultFunctionWithStatus` and `SlickMultipleResultFunctionWithStatus`
* Changed interface of `runWithStatus` and all dependant methods from `F[Either[StatusException, R]]` to `F[Either[StatusException, FunctionStatusWithData[R]]]` (and also their Seq-like versions)
* Consolidated common package types of the core/status and added type aliases
* Added `StatusAggregation` with three simple yet most likely used algorithms for aggregating error statuses, i.e. transformation from `F[Seq[Either[StatusException, FunctionStatusWithData[R]]]]` to `F[Either[StatusException, Seq[FunctionStatusWithData[R]]]]`
* Covered with plenty of unit and integration tests

Depends on #119 

Closes #118 
Closes #115


slick/src/main/scala/za/co/absa/fadb/slick/SlickQuery.scala
```diff
@@ -61,19 +60,26 @@ class SlickQueryWithStatus[R](
    */
   override def toStatusExceptionOrData(
     statusWithData: FunctionStatusWithData[PositionedResult]
-  ): Either[StatusException, R] = {
-    checkStatus(statusWithData) match {
-      case Left(statusException) => Left(statusException)
-      case Right(value)          => Right(getResult(value))
+  ): ExceptionOrStatusWithDataRow[R] = {
+    val o = checkStatus(statusWithData)
```
salamonpavel, line 64: Consider using more descriptive names for variables in this method.

core/src/main/scala/za/co/absa/fadb/status/aggregation/implementations/ByMajorityErrorsStatusAggregator.scala
```diff
@@ -0,0 +1,58 @@
...
+
+import za.co.absa.fadb.status.aggregation.StatusAggregator
+import za.co.absa.fadb.status.{ExceptionOrStatusWithDataResultAgg, ExceptionOrStatusWithDataRow}
+
+/**
+  *  `ByMajorityErrorsStatusAggregator` is a trait that extends the `StatusAggregator` interface.
+  *  It provides an implementation for aggregating error statuses of a function invocation into a single error
+  *  by choosing the error that occurred the most.
+  */
+trait ByMajorityErrorsStatusAggregator extends StatusAggregator {
```
salamonpavel, line 27: I don't know if this is needed. I can hardly imagine someone would want to aggregate errors in this way.

core/src/main/scala/za/co/absa/fadb/status/aggregation/implementations/ByFirstRowStatusAggregator.scala
```diff
@@ -0,0 +1,44 @@
...
+  *  It provides an implementation for aggregating error statuses of a function invocation into a single error
+  *  by choosing the first row that was returned to be the representative one
+  *  (i.e. if there is an error on row two or later, it would be ignored).
+  */
+trait ByFirstRowStatusAggregator extends StatusAggregator {
+
+  override def aggregate[R](statusesWithData: Seq[ExceptionOrStatusWithDataRow[R]]): ExceptionOrStatusWithDataResultAgg[R] = {
+    val firstRow = statusesWithData.headOption
+
+    val dataFinal = gatherDataWithStatuses(statusesWithData)
```
salamonpavel, line 33: We need to gather data only if first row didn't contain an error.

core/src/main/scala/za/co/absa/fadb/status/aggregation/implementations/ByFirstErrorStatusAggregator.scala
```diff
@@ -0,0 +1,41 @@
...
+/**
+  *  `ByFirstErrorStatusAggregator` is a trait that extends the `StatusAggregator` interface.
+  *  It provides an implementation for aggregating error statuses of a function invocation into a single error
+  *  by choosing the first error encountered to be the representative one (i.e. if there are multiple errors of other
+  *  types being returned, only the first one would be chosen and the rest would be ignored).
+  */
+trait ByFirstErrorStatusAggregator extends StatusAggregator {
+
+  override def aggregate[R](statusesWithData: Seq[ExceptionOrStatusWithDataRow[R]]): ExceptionOrStatusWithDataResultAgg[R] = {
+    val firstError = gatherExceptions(statusesWithData).headOption
```
salamonpavel, line 31: Inefficient.

core/src/main/scala/za/co/absa/fadb/status/aggregation/StatusAggregator.scala
```diff
@@ -0,0 +1,56 @@
...
+import za.co.absa.fadb.status.{ExceptionOrStatusWithDataResultAgg, ExceptionOrStatusWithDataRow, FunctionStatusWithData}
+
+/**
+  *  `StatusAggregator` is a base trait that defines the interface for aggregating the error statuses of a function
+  *  invocation. It provides methods to aggregate the error statuses into a single status information - this is
+  *  typically needed for database functions that retrieve multiple records.
+  */
+trait StatusAggregator {
+
+  private [aggregation] def gatherExceptions[R](
```
salamonpavel, line 29: We don't need to iterate over the whole sequence for certain aggregation approaches.

slick/src/main/scala/za/co/absa/fadb/slick/SlickQuery.scala
lsulak: Oh, damned! Apologies, this is a result of me debugging this and not changing it back. Rookie mistake :D

core/src/main/scala/za/co/absa/fadb/status/aggregation/implementations/ByFirstErrorStatusAggregator.scala
lsulak: True, I'll change this

core/src/main/scala/za/co/absa/fadb/status/aggregation/implementations/ByFirstRowStatusAggregator.scala
lsulak: Will change

core/src/main/scala/za/co/absa/fadb/status/aggregation/StatusAggregator.scala
lsulak: Yeah in most algorithms we won't, but for some perhaps we'll need, so I kept it here for now

benedeki:
This item depends on:
* AbsaOSS/fa-db#119
By **[Dependent Issues](https://github.com/z0al/dependent-issues)**.

core/src/main/scala/za/co/absa/fadb/Query.scala
```diff
@@ -26,9 +26,9 @@ trait Query[R]
 
 /**
  *  The basis for all query types of [[DBEngine]] implementations with status
- *  @tparam A - the initial result type of the query
- *  @tparam B - the intermediate result type of the query
- *  @tparam R - the final return type of the query
+ *  @tparam A - the initial result type of the query (a row basically, having status-related columns as well)
+ *  @tparam B - the intermediate result type of the query (a row without status columns, i.e. data only)
+ *  @tparam R - the final return type of the query (final version of B, depending on the needs, might be the same as B)
```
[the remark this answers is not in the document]
lsulak: Ok, will be changed

core/src/main/scala/za/co/absa/fadb/Query.scala
```diff
@@ -44,14 +44,15 @@ trait QueryWithStatus[A, B, R] {
...
    *  @param initialResult - the initial result of the query
    *  @return the result of the query or a status exception
    */
-  def getResultOrException(initialResult: A): Either[StatusException, R] = toStatusExceptionOrData(
-    processStatus(initialResult)
-  )
+  def getResultOrException(initialResult: A): ExceptionOrStatusWithDataRow[R] =
+    toStatusExceptionOrData(
+      processStatus(initialResult)
+    )
```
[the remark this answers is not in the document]
lsulak: Ok, will be changed

benedeki:
This item depends on:
* ~~AbsaOSS/fa-db#119~~
By **[Dependent Issues](https://github.com/z0al/dependent-issues)**.

core/src/main/scala/za/co/absa/fadb/status/package.scala
```diff
@@ -0,0 +1,34 @@
...
+  type ExceptionOrStatusWithDataRow[R] = Either[StatusException, FunctionStatusWithData[R]]
+
+  /**
+    * This is a representation of multiple rows returned from a DB function with processed status information,
+    * with error statuses aggregated to a single one.
+    *
+    * Note: R here represents a single row reduced by status-related columns, i.e. a type of data.
+    */
+  type ExceptionOrStatusWithDataResultAgg[R] = Either[StatusException, Seq[FunctionStatusWithData[R]]]
+}
```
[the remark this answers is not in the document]
lsulak: improved, thanks for the suggestion(s)!

core/src/main/scala/za/co/absa/fadb/status/package.scala
```diff
@@ -0,0 +1,34 @@
...
+    */
+  case class FunctionStatus(statusCode: Int, statusText: String)
+
+  /**
+    *  Represents a function status with data.
+    *  @param functionStatus the function status
+    *  @param data the data of one row (barring the status fields)
+    *  @tparam R the type of the data
+    */
+  case class FunctionStatusWithData[R](functionStatus: FunctionStatus, data: R)
```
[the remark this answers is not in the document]
lsulak: improved, thanks!

core/src/main/scala/za/co/absa/fadb/status/package.scala
```diff
@@ -0,0 +1,50 @@
...
+    */
+  case class FunctionStatus(statusCode: Int, statusText: String)
+
+  /**
+    *  Represents a function status with data.
+    *  @param functionStatus the function status
+    *  @param data the data of one row (barring the status fields)
+    *  @tparam D the type of the data
+    */
+  case class Row[D](functionStatus: FunctionStatus, data: D)
```
salamonpavel, line 34: I am not sure if `Row` is a good name. Also why `D` for the type parameter?

core/src/main/scala/za/co/absa/fadb/status/package.scala
```diff
@@ -0,0 +1,50 @@
...
+
+  /**
+    *  This is a representation of a single row returned from a DB function with processed status information.
+    *
+    *  Note: D here represents a single row reduced by status-related columns, i.e. a type of data.
+    */
+  type FailedOrRow[D] = Either[StatusException, Row[D]]
```
salamonpavel, line 41: How about `StatusExceptionOr[A]` - it's more common to have sth like 'ErrorOr[A]' or sth like my suggestion which has the type encoding the successful value in the name.

core/src/main/scala/za/co/absa/fadb/status/package.scala
```diff
@@ -0,0 +1,50 @@
...
+
+  /**
+    * This is a representation of multiple rows returned from a DB function with processed status information,
+    * with error statuses aggregated to a single one.
+    *
+    * Note: D here represents a single row reduced by status-related columns, i.e. a type of data.
+    */
+  type FailedOrRowSet[D] = Either[StatusException, Seq[Row[D]]]
```
salamonpavel, line 49: Same as above. I would reconsider the name.

core/src/main/scala/za/co/absa/fadb/DBFunction.scala
```diff
@@ -206,4 +222,85 @@ object DBFunction {
...
+  /**
+    *  `DBMultipleResultFunctionWithStatus` is an abstract class that represents a database function returning
+    *  multiple results with status information.
+    *  It extends the [[DBFunctionWithStatus]] class and overrides the apply method to return a sequence of results.
+    */
+  abstract class DBMultipleResultFunctionWithStatus[I, R, E <: DBEngine[F], F[_]](
+    functionNameOverride: Option[String] = None
+  )(implicit schema: DBSchema, dBEngine: E)
+    extends DBFunctionWithStatus[I, R, E, F](functionNameOverride)
+      with StatusAggregator {
```
salamonpavel, line 235: I don't like the fact it requires `StatusAggregator` to be mixed-in. I believe this should be optional and the return value of `DBMultipleResultFunctionWithStatus` should be a sequence of `Either` (or it's alias). The aggregation should be in my opinion applied optionally later on as one might be interested also in all statuses.

slick/src/main/scala/za/co/absa/fadb/slick/SlickQuery.scala
```diff
@@ -60,20 +59,26 @@ class SlickQueryWithStatus[R](
...
-  ): Either[StatusException, R] = {
+    statusWithData: Row[PositionedResult]
+  ): FailedOrRow[R] = {
     checkStatus(statusWithData) match {
-      case Left(statusException) => Left(statusException)
-      case Right(value)          => Right(getResult(value))
+      case Left(statusException)  => Left(statusException)
+      case Right(value) =>
+        val s = value.functionStatus
+        val d = getResult(value.data)
```
salamonpavel, line 68: Please rename to `data`

core/src/main/scala/za/co/absa/fadb/status/package.scala
lsulak: `D` simply says that it's the `data`, or, 'effective/actual data' (sometimes this term can be find in literature). So basically a row without the status-related columns.

Hmm, why not? `Row` indicates that 1 instance of this should represent 1 row being returned from the DB. I had the name much more complicated, and it's true that `Row` is pretty simple, naive even, but when I checked the usecases and thought about its internal structure, I think that it's good enough name for this (couldn't come up with anything better and I didn't want the names to be sentences :D )

Also, this naming was not chosen to be 'isolated' to this `Row` class - we wanted to make it kind of relatable and consistent with the rest of the case classes in this module - see the classes below in this module :)  

But as always, I'm open to proposals

slick/src/main/scala/za/co/absa/fadb/slick/SlickQuery.scala
lsulak: Okay, no problem with that

slick/src/main/scala/za/co/absa/fadb/slick/SlickQuery.scala
```diff
@@ -60,20 +59,26 @@ class SlickQueryWithStatus[R](
...
-    statusWithData: FunctionStatusWithData[PositionedResult]
-  ): Either[StatusException, R] = {
+    statusWithData: Row[PositionedResult]
+  ): FailedOrRow[R] = {
     checkStatus(statusWithData) match {
-      case Left(statusException) => Left(statusException)
-      case Right(value)          => Right(getResult(value))
+      case Left(statusException)  => Left(statusException)
+      case Right(value) =>
+        val s = value.functionStatus
```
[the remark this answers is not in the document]
lsulak: okay,  I have no problem with that

core/src/main/scala/za/co/absa/fadb/DBFunction.scala
lsulak: yes that's basically the hardest 'design decision' in this PR I think. I appreciate / see the benefits of both approaches, that's why I was kind of undecided yet and 'waiting' for 3rd opinion :D

core/src/main/scala/za/co/absa/fadb/status/package.scala
lsulak: Hmm but sometimes it can happen that the type parameter is provided somehow automatically (is the compiler can understand it from the caller, then I won't need to specify it explicitly), which would lead to odd name

On the other hand, we can always explicitly supply it..

I must admit that I haven't really seen such approach in the past, but then you are much more experienced in Scala - I'll think about it, because the main idea was to make those 3 classes relatable = so I value 3 classes named well enough and relate-able, rather than each class perfectly named but not relate-able. WDYT?

core/src/main/scala/za/co/absa/fadb/status/package.scala
lsulak: acknowledged, but see my answer above

benedeki: I would also consider name change, but little different.
`FailedOrRows[D]` or `FailedOrRowSeq[D]`? 🤔

lsulak: I like `FailedOrRows`, thanks

core/src/main/scala/za/co/absa/fadb/Query.scala
```diff
@@ -26,9 +26,9 @@ trait Query[R]
 
 /**
  *  The basis for all query types of [[DBEngine]] implementations with status
- *  @tparam A - the initial result type of the query
- *  @tparam B - the intermediate result type of the query
- *  @tparam R - the final return type of the query
+ *  @tparam A - the initial result type of the query (a row basically, having status-related columns as well)
+ *  @tparam B - the intermediate result type of the query (a row without status columns, i.e. data only)
```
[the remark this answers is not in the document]
lsulak: adopted, I like it!

core/src/main/scala/za/co/absa/fadb/Query.scala
```diff
@@ -37,21 +37,22 @@ trait QueryWithStatus[A, B, R] {
    *  @param initialResult - the initial result of the query
    *  @return the status with data
```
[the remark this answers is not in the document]
lsulak: sure, changed on all the places

benedeki (approved):
Definitely an improvement!

core/src/main/scala/za/co/absa/fadb/DBFunction.scala
lsulak: Okay I though about it and:
* I think status aggregation in general is gonna be used quite common and is therefore very helpful - just need to provide implementation that is convenient for us and users
* I still really like and prefer mix-in approach - it's consistent and 'expected' from user point of view how things are plugged together, like nice reusable and changeable lego blocks
* I understand that having a mandatory agg implementation might be too much and sometimes not needed. Therefore I implemented it separately - one with mixin and the other without.
* With this 'new' approach, I created respective 'utils' ('helper' classes) to both, Doobie and Slick, and sufficiently covered it all with tests.

Does that influence your opinion on this status aggregation related feature?

benedeki (approved):
LGTM

salamonpavel (approved):

PR #134, opened 2024-08-15
#133: don't handle data part on error Status code query results

### LINKED ISSUE #133: Error Status code means data can be null
## Background
Error Status Code usually means that the data for such row is all NULL. In such case, a return type must be an `Option` and must be specified by a user. Ideally we should not burden our users with such cases and we should programmatically ensure that when the DB query returns error-like status codes / results, data is not taken into account at all.

There might be some `Read` related changes for Doobie, especially because we'll be introducing an Option return type for data. This might be a good read:
* https://blog.rockthejvm.com/doobie/
* https://stackoverflow.com/questions/58530983/doobie-cannot-find-or-construct-a-read-instance-for-type-t

lsulak:
Closes #133


lsulak:
Closing in favor of https://github.com/AbsaOSS/fa-db/pull/137

PR #137, opened 2025-02-10
#133 Fix doobie error when status is not OK and data is missing

### LINKED ISSUE #133: Error Status code means data can be null

jakipatryk:
Fixes #133

### DIFF
--- a/doobie/src/main/scala/za/co/absa/db/fadb/doobie/StatusWithData.scala
+++ b/doobie/src/main/scala/za/co/absa/db/fadb/doobie/StatusWithData.scala
@@ -16,12 +16,30 @@
 
 package za.co.absa.db.fadb.doobie
 
+import doobie.Read
+
 /**
- *  Represents a function status with data (basically a row returned from a DB).
- *
- *  Note: data here represent one row that has status-related fields omitted. In some scenarios, it actually might
- *    be missing - i.e. query returns only status information with no data - pretty common scenario when a DB
- *    function returns error, e.g.: (41, 'NoDataFound', NULL). In this case, R must be an Option - to be specified
- *    by a user.
+ * Represents a function status with data (basically a row returned from a DB).
  */
-case class StatusWithData[R](status: Int, statusText: String, data: R)
+case class StatusWithData[R](status: Int, statusText: String, data: Option[R])
+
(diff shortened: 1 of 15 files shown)


doobie/src/main/scala/za/co/absa/db/fadb/doobie/StatusWithData.scala
```diff
@@ -16,12 +16,30 @@
...
+  // This is for backward compatibility of #133 (which makes the data optional by default when reading from DB),
+  // as some users were overcoming the bug by wrapping the result type in Option themselves.
+  // Doobie provides Read for Option[T] for most of the types out of the box, except when T itself is an Option.
+  implicit def readOptionToReadDoubleOption[T](
+                                                implicit r: Read[StatusWithData[T]]
+                                              ): Read[StatusWithData[Option[T]]] = {
+    r.map {
+      case statusWithData @ StatusWithData(_, _, Some(data)) => statusWithData.copy(data = Some(Some(data)))
+      case statusWithData @ StatusWithData(_, _, None) => statusWithData.copy(data = None)
+    }
```
lsulak, line 42: this segment is very nice, including the doc and backward compatibility big - thank you!

lsulak (approved):
I really couldn't find a single thing, not even minor / naming. Really good job, you have a great ability to understand and involve yourself into anything & quickly.

Perhaps wait for @salamonpavel and/or @benedeki for some further review/opinions, but basically something like this I had in my mind, for some reasons I found a bit more complicated way and didn't finish it :D

salamonpavel (approved):
Good job. Thank you.

benedeki (approved):
Nice work, thank you.

PR #144, opened 2025-09-17
#142: Add the ability to list class constructor parameter names

### LINKED ISSUE #142: Add the ability to list class constructor parameter names
## Background
The _Fa-db_ object needs the fields to select from the appropriate function. Usually they are equivalent to a constructor parameters of a class that will hold the returned data.

## Feature
Add the ability to easilly extract a class/case class constructor parameter names. Utilize `NamingConvnetion`.
It can be a class, a trait or a function in an object.

benedeki:
Closes #142 

## Summary by CodeRabbit

* **New Features**
  * Added a utility to extract constructor field names with configurable naming conventions and input validation for unsupported types.

* **Tests**
  * Added unit tests covering default, explicit, and implicit naming behaviors and expected failure scenarios.

* **Chores**
  * Expanded ignore list for editor/build tooling.

* **Dependencies**
  * Added scala-reflect to project dependencies.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/core/src/main/scala/za/co/absa/db/fadb/utils/ClassFieldNamesExtractor.scala
+++ b/core/src/main/scala/za/co/absa/db/fadb/utils/ClassFieldNamesExtractor.scala
@@ -51,7 +51,7 @@ object ClassFieldNamesExtractor {
    * @tparam T               - type to investigate and extract field names from
    * @return                 - list of field names
    */
-  def extract[T: TypeTag]()(
+  def extract[T: TypeTag](
                            implicit namingConvention: NamingConvention = SnakeCaseNaming.Implicits.namingConvention
                          ): Seq[String] = {
     doExtract[T](namingConvention)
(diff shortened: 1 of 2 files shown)


lsulak (commented):
The formatting is a bit wild, branch out of base, license header is missing, and the compilation process is failing, but none is probably very difficult to fix; I'm happy to approve then

core/src/main/scala/za/co/absa/db/fadb/utils/ClassFieldNamesExtractor.scala
```diff
@@ -26,28 +26,23 @@ object ClassFieldNamesExtractor {
...
       throw new IllegalArgumentException(s"${tpe.typeSymbol} is not a case class nor a class")
     }
+    val cl = tpe.typeSymbol.asClass
+    if (cl.isPrimitive) {
+      throw new IllegalArgumentException(s"${tpe.typeSymbol} is a primitive type, extraction is not supported")
+    }
+    if (cl.isTrait) {
+      throw new IllegalArgumentException(s"${tpe.typeSymbol} is a trait, extraction is not supported")
+    }
+    tpe
```
lsulak: I like the changes! I learned this technique when I read Clean Code book for the first time a while ago.

However, it's not equivalent, now this one is missing: `if (cl.isCaseClass || cl.isClass)` - probably only CaseClass, i.e. the first condition, is what we still need here? The second is probably addressed on few lines above, on the first condition actually (line 20)?

PR #160, opened 2026-05-28
multiple function calls within single transaction

salamonpavel:
Closes #159 
Adds support for multiple function calls within single transaction in Doobie module.

## Summary by CodeRabbit

* **New Features**
  * Compose multiple database operations into a single atomic transaction, including a utility to run connection-level DB actions within that transaction.

* **Tests**
  * Added integration tests covering multi-operation transactions, rollback on failure, and backward-compatibility of direct execution.

### DIFF
--- a/doobie/src/main/scala/za/co/absa/db/fadb/doobie/DoobieEngine.scala
+++ b/doobie/src/main/scala/za/co/absa/db/fadb/doobie/DoobieEngine.scala
@@ -99,3 +99,21 @@ class DoobieEngine[F[_]: Async](val transactor: Transactor[F]) extends DBEngine[
   def runConnectionIO[R](cio: ConnectionIO[R]): F[R] =
     cio.transact(transactor)
 }
+
+object DoobieEngine {
+
+  /**
+   *  Executes an arbitrary `ConnectionIO` program in a single transaction,
+   *  using an implicit `DoobieEngine` instance.
+   *  Useful in repository classes where the engine is not directly available
+   *  but is in implicit scope.
+   *
+   *  @param cio the `ConnectionIO` program to execute
+   *  @param engine the `DoobieEngine` instance (typically implicit)
+   *  @tparam R the result type of the program
+   *  @tparam F the effect type
+   *  @return the result wrapped in the effect type `F`
+   */
+  def runConnectionIO[R, F[_]: Async](cio: ConnectionIO[R])(implicit engine: DoobieEngine[F]): F[R] =
+    engine.runConnectionIO(cio)
+}
--- a/doobie/src/main/scala/za/co/absa/db/fadb/doobie/DoobieEngine.scala
+++ b/doobie/src/main/scala/za/co/absa/db/fadb/doobie/DoobieEngine.scala
@@ -86,4 +86,16 @@ class DoobieEngine[F[_]: Async](val transactor: Transactor[F]) extends DBEngine[
   override def runWithStatus[R](query: QueryWithStatusType[R]): F[Seq[FailedOrRow[R]]] = {
     executeQueryWithStatus(query)(query.readStatusWithData)
   }
+
+  /**
+   *  Executes an arbitrary `ConnectionIO` program in a single transaction.
+   *  This enables composing multiple database operations (including multiple fa-db function calls)
+   *  into a single atomic transaction.
+   *
+   *  @param cio the `ConnectionIO` program to execute
+   *  @tparam R the result type of the program
+   *  @return the result wrapped in the effect type `F`
+   */
+  def runConnectionIO[R](cio: ConnectionIO[R]): F[R] =
+    cio.transact(transactor)
 }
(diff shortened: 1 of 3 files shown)


doobie/src/main/scala/za/co/absa/db/fadb/doobie/DoobieFunction.scala
```diff
@@ -119,6 +120,19 @@ trait DoobieFunction[I, R, F[_]] extends DoobieFunctionBase[R] {
...
+   *  This enables composing multiple function calls into a single transaction using for-comprehensions
+   *  over `ConnectionIO`, and then executing them atomically via `DoobieEngine.runConnectionIO`.
+   *
+   *  @param values the input values for the function
+   *  @return the `ConnectionIO[Seq[R]]` representing the database function call
+   */
+  def toConnectionIO(values: I): ConnectionIO[Seq[R]] = {
+    val fragments = toFragmentsSeq(values)
+    composeFragments(fragments).query[R].to[Seq]
+  }
```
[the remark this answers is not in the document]
jakipatryk: I think this one is worth doing

jakipatryk (commented):
Makes sense, the comment from CodeRabbit is probably worth addressing though.

doobie/src/main/scala/za/co/absa/db/fadb/doobie/DoobieEngine.scala
```diff
@@ -86,4 +86,34 @@ class DoobieEngine[F[_]: Async](val transactor: Transactor[F]) extends DBEngine[
...
+  /**
+   *  Executes an arbitrary `ConnectionIO` program in a single transaction.
+   *  This enables composing multiple database operations (including multiple fa-db function calls)
+   *  into a single atomic transaction.
+   *
+   *  @param cio the `ConnectionIO` program to execute
+   *  @tparam R the result type of the program
+   *  @return the result wrapped in the effect type `F`
+   */
+  def runConnectionIO[R](cio: ConnectionIO[R]): F[R] =
```
benedeki, line 99: The method name doesn't sound too idiomatic to me. But at the end it might be a good thing, in regard of my final comment (coming 😉 )

doobie/src/test/scala/za/co/absa/db/fadb/doobie/DoobieTransactionCompositionIntegrationTests.scala
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  test("Compose multiple creates in a single transaction") {
+    val timestamp = System.currentTimeMillis()
+    val lastName = s"BatchTest_$timestamp"
+
+    val program: ConnectionIO[Unit] = for {
+      _ <- createActor.toConnectionIOSingle(CreateActorRequestBody("BatchActor1", lastName))
+      _ <- createActor.toConnectionIOSingle(CreateActorRequestBody("BatchActor2", lastName))
+      _ <- createActor.toConnectionIOSingle(CreateActorRequestBody("BatchActor3", lastName))
```
benedeki, line 101: What would happen if a "old fashion" call to apply is in the middle here. E.g.
```scala
      _ <- createActor.toConnectionIOSingle(CreateActorRequestBody("BatchActor2", lastName))
     val foo = createActor(CreateActorRequestBody("Intruder1", uniqueLast)).unsafeRunSync()
     bar <= createActor(CreateActorRequestBody("Intruder2", uniqueLast)).unsafeRunSync()
      _ <- createActor.toConnectionIOSingle(CreateActorRequestBody("BatchActor3", lastName))
```

benedeki (commented):
The code looks elegant, indeed 👏 

The problem I have with the introduction of transaction is, that it goes against the basic paradigm of the library.
_Separation of duties and responsibilities between the database and the application_
Particularly you rely on the fact that on the database side, there is a single instance supporting transaction. Or conversely you enforce the fact (of single instance) preventing to migrate the database from single instance to cluster.

So from that point of view the less idiomatic method name serves well, as it prevents thoughtless use to some extent.
Speaking off, I am not sure in what state id the documentation, but this method surely wasn't added there (if any exists).

doobie/src/test/scala/za/co/absa/db/fadb/doobie/DoobieTransactionCompositionIntegrationTests.scala
```diff
@@ -0,0 +1,140 @@
...
+class DoobieTransactionCompositionIntegrationTests extends AnyFunSuite with DoobieTest {
+
+  private val engine = new DoobieEngine(transactor)
+
+  class CreateActor(implicit schema: DBSchema, dbEngine: DoobieEngine[IO])
+    extends DoobieSingleResultFunctionWithStatus[CreateActorRequestBody, Int, IO](
+      values => Seq(fr"${values.firstName}", fr"${values.lastName}")
+    )
+    with StandardStatusHandling {
+    override def fieldsToSelect: Seq[String] = super.fieldsToSelect ++ Seq("o_actor_id")
```
lsulak, line 40: ah damn, I see it's here:  https://github.com/AbsaOSS/fa-db/blob/master/demo_database/src/main/postgres/integration/V1.2.5__create_actor.sql#L39

but basically the convention we made a while ago was `i_` prefix for input but no prefix for output parameters

